PR #19, opened 2024-03-03
Updated install command

DawidPietrykowski:
Updated the install command in README.md as cargo is capable of cloning the repository itself so it's easier for users


altsem (approved):
Was unaware of this one, amazing! :+1:

Issue #50, opened 2024-03-11
Publish release binaries on release


chasinglogic:
It would be very helpful if there were binaries available for download so you don't need to install the Rust toolchain to install gitu.

I'd be happy to submit a PR for this as I've gone through the process a few times with my own projects if this would be accepted

PR #52, opened 2024-03-11
Add release automation which publishes release binaries

chasinglogic:
This adds a continuous delivery pipeline which will run on release tags.
It currently will build a release binary for Linux x86_64, Linux ARM,
and MacOS x86_64 (which can be run on M* series macs).

It has code in it already to support cross compiling using
[cross](https://github.com/cross-rs/cross) which can be setup later.
I've yet to get it working flawlessly on my own projects so I have left
the setup code in but not utilised it.

This fixes https://github.com/altsem/gitu/issues/50


chasinglogic:
You'll need to setup a github token and add it to the project secrets for this to work. You can find instructions for that here and here:
- https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-fine-grained-personal-access-token
- https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository

.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+      GITU_VERSION: ""
+    outputs:
+      upload_url: ${{ steps.release.outputs.upload_url }}
+      gitu_version: ${{ env.GITU_VERSION }}
+    steps:
+      - name: Get the release version from the tag
+        shell: bash
+        if: env.GITU_VERSION == ''
+        run: |
+          # See: https://github.community/t5/GitHub-Actions/How-to-get-just-the-tag-name/m-p/32167/highlight/true#M1027
```
altsem, line 21: The link is broken, seems there's a `github.ref_name` / `GITHUB_REF_NAME` nowadays: https://github.com/orgs/community/discussions/26686

.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+          rust: stable
+          target: arm-unknown-linux-gnueabihf
+        - build: macos
+          os: macos-latest
+          rust: stable
+          target: x86_64-apple-darwin
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
```
altsem, line 60:
```suggestion
      uses: actions/checkout@v4
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: ${{ matrix.rust }}
+        target: ${{ matrix.target }}
```
altsem, line 66: cargo / rust is already available when using ubuntu-latest (pre-installed for actions I believe?)

README.md
```diff
@@ -40,6 +40,29 @@ Run the command (recommended):
...
+platform=$(uname -s)
+arch=$(uname -m)
+# If you're running on an M1 Macbook run this:
+# arch="x86_64"
+download_url=$(curl -s https://api.github.com/repos/altsem/gitu/releases/latest | grep "browser_download_url.*$arch.*${platform,,}" | cut -d : -f 2,3 | sed 's/"//g' | xargs)
+curl -L -o /tmp/gitu.tar.gz "$download_url"
+tar -C /tmp -xzvf /tmp/gitu.tar.gz
+mv $(find /$(readlink /tmp) -perm +111 -type f -name gitu 2>/dev/null) /usr/local/bin/
+```
+
```
altsem, line 65: Perhaps this part about installing using a script is superfluous. I'd love to get publishing to package managers in the future though, like Homebrew! I'd keep the note about the `PATH` and `/usr/local/bin`

.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+      shell: bash
+      run: |
+        cargo install cross
+        echo "CARGO=cross" >> $GITHUB_ENV
+
+    - name: Show command used for Cargo
+      run: |
+        echo "cargo command is: ${{ env.CARGO }}"
+        echo "target flag is: ${{ env.TARGET_FLAGS }}"
+        echo "target dir is: ${{ env.TARGET_DIR }}"
```
altsem, line 78: Is this required? Else just invoking e.g. `cross build` would be self-explanatory.

altsem (changes requested):
Thanks for the work put in here! I had some feedback, see if they make sense :)

altsem:
> You'll need to setup a github token and add it to the project secrets for this to work. You can find instructions for that here and here:
> 
>     * https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-fine-grained-personal-access-token
> 
>     * https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository

Actually GITHUB_TOKEN is set automatically, I think it already has access to the release api. Guess it's easy to find out by trying.

.github/workflows/release.yml
chasinglogic: This uses a matrix to allow using different versions of rust. I've only added stable binaries but you could easily duplicate the targets to build nightly versions as well.

README.md
chasinglogic: Removed

.github/workflows/release.yml
chasinglogic: I always preferred having as much info as possible when debugging these builds. I have removed it

.github/workflows/release.yml
chasinglogic: Also I couldn't find any references to cargo or rust in https://github.com/actions/runner-images/blob/main/images/ubuntu/toolsets/toolset-2204.json

.github/workflows/release.yml
chasinglogic: updated, I've omitted the community discussion link since they don't seem to stick around

.github/workflows/release.yml
altsem: No biggie, lets stick with it :+1:

altsem (approved):
I'll give it a shot, thanks again! :)

PR #56, opened 2024-03-12
Adds Nix flake for building on NixOS

### LINKED ISSUE #54: Nix flake and/or package
Any interest in packaging this in a Nix flake and/or package? It would make it much easier than toolchaining Rust.

t56k:
Closes #54 (ignoring macOS for the time being--my Nix install in that env is borked). Might need some extra `buildInputs` for macOS/Darwin, happy to amend when I can.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,46 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    crane.url = "github:ipetkov/crane";
+    crane.inputs.nixpkgs.follows = "nixpkgs";
+
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs = {
+        nixpkgs.follows = "nixpkgs";
+        flake-utils.follows = "flake-utils";
+      };
+    };
+
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, crane, rust-overlay, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        craneLib = crane.lib.${system};
+
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [ (import rust-overlay) ];
+        };
+
+        gitu = craneLib.buildPackage {
+          src = craneLib.cleanCargoSource (craneLib.path ./.);
+          doCheck = false;
+          buildInputs = with pkgs; [ openssl ];
+          nativeBuildInputs = with pkgs; [ pkg-config ];
+        };
+      in {
+        checks = { inherit gitu; };
+
+        packages.default = gitu;
+        apps.default = flake-utils.lib.mkApp { drv = gitu; };
+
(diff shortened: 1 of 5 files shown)


Mojken:
Nice work! Please update the readme with install instructions as well :)

postsolar:
Just for my own learning, why is `crane` needed given that you can just give `rustPlatform.buildRustPackage` the local lockfile?

altsem:
I'll leave this here just for a day or two to gather input, as I'm not adept with nix myself.
A mention in the README would be nice!

t56k:
> Just for my own learning, why is `crane` needed given that you can just give `rustPlatform.buildRustPackage` the local lockfile?

[Crane](https://crane.dev/) is a relatively-simplified way of making a temporary Rust environment on Nix platforms with platform-specific requirements, checks, and the like, so that `gitu`'s builds are (more) reproducible. Helix's [example](https://github.com/helix-editor/helix/blob/master/flake.nix) shows a more comprehensive use-case. In my experience it just makes building easier, but if there's enough pushback I'm happy to change it. Also it's what I've used in the past and what has worked for my devops deployments.

evanrichter:
gitu is already packaged in the official nixpkgs repo: https://github.com/NixOS/nixpkgs/blob/ddbd484a31481757a93d1ce7122dcb8469ff5fb2/pkgs/by-name/gi/gitu/package.nix

this will provide a binary cache, and it works on both macOS and linux

this flake does provide a devShell, which would be useful for local development if maintainers want to lock their development dependencies

flake.nix
```diff
@@ -0,0 +1,49 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    crane.url = "github:ipetkov/crane";
+    crane.inputs.nixpkgs.follows = "nixpkgs";
+
+    rust-overlay = {
```
justinrubek, line 7: Could we consider using [fenix](https://github.com/nix-community/fenix/tree/main) instead? rust-overlay ends up being larger and I find it much more cumbersome to use

flake.nix
```diff
@@ -0,0 +1,49 @@
...
+        nixpkgs.follows = "nixpkgs";
+        flake-utils.follows = "flake-utils";
+      };
+    };
+
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, crane, rust-overlay, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
```
justinrubek, line 19: I think it'd be cleaner to avoid `flake-utils` and be explicit about what systems are supported. Preferably replacing this with [flake-parts](https://github.com/hercules-ci/flake-parts)

https://ayats.org/blog/no-flake-utils/

flake.nix
t56k: I've switched to fenix--good idea.

justinrubek (approved):
This looks great to me and it works! Thanks for accepting my input.

altsem (approved):

Issue #37, opened 2024-03-05
Stash commands and status


altsem:
We're missing the stash list on the status page, as well as stash commands.

Start with save / pop / apply.
Some of these commands open up a prompt, like `zz` prompts you to name the stash.
`zp` will also prompt which stash to pop (default to 0, or the marked one).

![image](https://github.com/altsem/gitu/assets/3618477/2ef5c971-988b-4ef8-bd23-5f9583ce1f79)

altsem:
I think keeping the reference to Repository immutable is not a requirement, but makes the code easier to reason about. If it needs to be a Rc<RefCell<Repository>> (I think?), it's not a huge deal.

Had a peek over at Gitoxide and it seems there's no support for it, else it could have been a good segue into starting to use that.

I found the `git` cli function [here](https://github.com/git/git/blob/945115026aa63df4ab849ab14a04da31623abece/builtin/stash.c#L171).
It seems to be referencing `refs/stash`, perhaps there's a way to list these with more nitty-gritty git2 commands?

altsem:
@golden-expiriensu I played around with it a bit, stole some code from the log fn. It lists a lot more than just what `git stash` would, not sure why. But the stash entries are indeed found in there.

```
#[test]
fn stash_list() {
    let repo = Repository::open(".").unwrap();
    let mut revwalk = repo.revwalk().unwrap();
    revwalk.push_ref("refs/stash").unwrap();
    dbg!(revwalk
        .flat_map(|maybe_oid| maybe_oid.and_then(|oid| repo.find_commit(oid)))
        .map(|commit| commit.summary().unwrap().to_string())
        .collect::<Vec<_>>());

    todo!();
}
```

Again, if `git_stash_foreach` proves to be really simple to use, we could compromise and get rid of the immutable reference to Repository!

golden-expiriensu:
I believe I found another way, under the hood git uses reflog to find changes of ref/stash file, so I think `git reflog stash` is a solution

PR #73, opened 2024-03-16
feat: stash status, save, pop, apply and drop

golden-expiriensu:
Almost everything is working as expected, but unfortunately I have no idea so far how to
- Successfully execute several commands per operation (for stash unstaged we need 3 commands). Current solution seems to not work :(
- Work with TargetOp. I suppose for passing an argument to pop/apply/drop I need to use TargetOp, because ops like stage/unstage apparently use it, but it is pretty confusing to me. I will be really grateful if you can explain how to work with it, and maybe we can also add some docs for ops for future contributors in this PR


golden-expiriensu:
Resolves https://github.com/altsem/gitu/issues/37 btw

altsem:
Yea, TargetOp was a mess, I've been trying to make this interface better. And now it's gone. Apologize for the conflicts :man_facepalming:.
I'll have a look and see if I can help :+1:

altsem:
This is how it looks like in master right now (except for the doc):
```rust
pub(crate) type Action = Rc<dyn FnMut(&mut State, &mut Term) -> Res<()>>;

pub(crate) trait OpTrait: Display {
    /// Get the implementation (which may or may not exist) of the Op given some TargetData.
    /// This indirection allows Gitu to show a contextual menu of applicable actions.
    fn get_action(&self, target: Option<&TargetData>) -> Option<Action>;

    /// This indicates whether the Op is meant to read and
    /// act on TargetData. Those are listed differently in the help menu.
    fn is_target_op(&self) -> bool {
        false
    }
}
```
image


Starting to feel the complexity vs the gains of doing this is kind of not worth it.

Perhaps should be changed back to just be a `trigger` function, and nothing more.
```rust
pub(crate) trait OpTrait: Display {
    fn trigger(&self, target: Option<&TargetData>) -> Action;
}
```

This means you would lose the contextual "target" column (here `ret Show` is listed with the other ops):
image

golden-expiriensu:
I am doing a rebase right now and actually new updates are looking pretty good so far, so its okay, no worries

golden-expiriensu:
@altsem Sorry for `stash_push_action_prompt_update` arguments, I played with lifetimes and lost :(

altsem:
Very cool! Having a look :)

src/ops/stash.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[derive(Default, Clone, Copy, PartialEq, Eq, Debug, Display)]
+#[display(fmt = "Stash working tree")]
+pub(crate) struct StashWorktree;
+impl OpTrait for StashWorktree {
+    fn get_action(&self, _target: Option<&TargetData>) -> Option<Action> {
+        // TODO: how to execute several commands?
+        // Ideally, for unstaged stashing we need to commit index, stash unstaged and then reset
+        // last commit. In the current implementation we include index in the stash despite index
+        // remains untouched.
```
altsem, line 33: I tried this in magit, here's what it seems to be doing:

stash worktree:
```
  0 git … update-ref --create-reflog -m hello refs/stash 85a1205d04c2c49ebaf6b905de844e96eddfe55d 1ef91d084e4a5e8e3d29ada319e52e50ea801706
  0 git … checkout -- .
  ```
  
stash index:
```
  0 git … update-ref --create-reflog -m hello refs/stash abe8de6c219ffa784d7fe038e8ee95e3b20a9
  0 git … apply --reverse --cached --ignore-space-change -
  0 git … apply --reverse --ignore-space-change -
```

It's still running multiple commands though.
It is of course possible to do, but it'd be nice if it jacks in to the existing interfaces and displays some output.

Comment (altsem) [reply src/ops/stash.rs:33]: Perhaps we should comment this out and leave the TODO? And remove the references to 'StashWorktree' for now too.

File: src/screen/status.rs [hunk 2]
```diff
@@ -249,6 +254,31 @@ fn create_status_section_items<'a>(
     .chain(items::create_diff_items(config, diff, &1, true))
 }
 
+fn create_stash_list_section_items<'a>(
```
altsem, line 257: It'd be nice if this header didn't show when there are no stashed changes!

altsem (changes requested):
Amazing work with this! The tests are failing, and perhaps it could use some new tests too (found at `tests/mod.rs`).

You can `cargo install cargo-insta` and `cargo test && cargo insta review`, in order to update the snapshots!

src/ops/stash.rs
golden-expiriensu: I not even sure why they run such complicated command, probably I don't understand something. I added running 3 commands, so the only issue right now is that we show only last run command, so I left TODO for that. It is working as expected so I think we can include that

src/screen/status.rs
golden-expiriensu: Yeah, this makes sense, I will try to do that

golden-expiriensu:
> [golden-expiriensu#1](https://github.com/golden-expiriensu/gitu/pull/1)

Ooh, thank you very much! I actually managed to solve this issue with `stash_push_action<const N: usize>(args: [&'static str; N])`. So pick which variant you would like to keep in the repo and I will update code accordingly

src/screen/status.rs
golden-expiriensu: Done

src/screen/status.rs
```diff
@@ -249,6 +254,36 @@ fn create_status_section_items<'a>(
...
+        vec![]
+    } else {
+        let style = &config.style;
+        vec![
+            Item {
+                display: Line::raw(""),
+                depth: 0,
+                unselectable: true,
+                ..Default::default()
+            },
```
altsem, line 273: This function could save some space. Saw that there's more of these in this module too.
```suggestion
            items::blank_line(),
```

src/ops/stash.rs
```diff
@@ -0,0 +1,181 @@
...
+use derive_more::Display;
+use std::{process::Command, rc::Rc};
+use tui_prompts::State as _;
+
+#[derive(Default, Clone, Copy, PartialEq, Eq, Debug, Display)]
+#[display(fmt = "Stash")]
+pub(crate) struct Stash;
+impl OpTrait for Stash {
+    fn get_action(&self, _target: Option<&TargetData>) -> Option<Action> {
+        Some(stash_push_action(["--include-untracked"]))
```
altsem, line 12: We don't have arguments yet, and this differs from git/magit (which is fine), but it might come as a surprise to some.
I'm thinking the `#[display(fmt = "Stash")]` could include this information for now. `"Stash (include untracked)`.
What do you think?

golden-expiriensu: This is very reasonable

golden-expiriensu: Screenshot 2567-03-18 at 14 00 02
I added this message only for 'z' stash because for other ones there is no enough space

altsem:
Looks good so far! Just some tests missing. Let me know if you need help with this, I wouldn't mind writing them either @golden-expiriensu.

Then when you feel it's ready, let me know and I'll have a final look / test things out manually too.

Haven't settled on a strict commit format, I'll update the PR name just in case I forget. The "feat: " prefix will result in a changelog entry.

golden-expiriensu:
Hey @altsem, I added tests for stash functionality, the PR is ready for the final review, let me know if I missed something
P.S. Btw, the tooling and test setup is awesome, I had a great time writing it

altsem:
@golden-expiriensu awesome :)

Tested it out quickly, there's an issue with StashWorktree though.
I had a pre-commit hook causing the commit to fail, and rest to mess up my state.

I saw there's this option --keep-index:
```sh
git stash --keep-index --include-untracked
```

Does this solve it all?

golden-expiriensu:
@altsem of course we can! I don' t know why I decided to use commits instead of stashes in the first place

golden-expiriensu:
@altsem done, it now uses stash as a temporary storage of index when stashing working tree, PTAL

altsem:
I read the code and had another go at testing it. Pretty happy with the result! Found some interesting differences between cli git and magit.
It seems stashing things with magit will "remember" which things are staged and not.

Gj! I'll merge this! :rocket:

altsem (approved):

PR #100, opened 2024-03-28
feat: show multiple command outputs in popup

altsem:
relates to: #2


golden-expiriensu:
> @golden-expiriensu This now works :)
> Think I found another edge-case as well
> ```
> $ git stash push --staged
> No staged changes
> $ git stash push --include-untracked
> 
> $ git stash pop 1
> fatal: log for 'refs/stash' only has 1 entries
> ```

Hmm, indeed, I'll fix that

PR #103, opened 2024-03-29
fix: stash working tree should work when everything/nothing is staged

golden-expiriensu:
Stash working tree is running 3 commands so there is some edge cases which we should take into account:
1. If everything is staged don't do anything
2. If nothing is staged don't try to save index in the stash and then later pop it (either errors or pops non-related stashes)

Addresses: https://github.com/altsem/gitu/pull/100#issuecomment-2025832482

### DIFF
--- a/src/ops/stash.rs
+++ b/src/ops/stash.rs
@@ -1,6 +1,7 @@
 use super::{Action, OpTrait};
 use crate::{items::TargetData, prompt::PromptData, state::State, term::Term, Res};
 use derive_more::Display;
+use git2::{Repository, Status, StatusOptions};
 use std::{process::Command, rc::Rc};
 use tui_prompts::State as _;
 
@@ -33,15 +34,21 @@ impl OpTrait for StashWorktree {
     fn get_action(&self, _target: Option<&TargetData>) -> Option<Action> {
         let update_fn = move |state: &mut State, term: &mut Term| -> Res<()> {
             if state.prompt.state.status().is_done() {
+                if is_working_tree_empty(&state.repo)? {
+                    state.prompt.reset(term)?;
+                    return Ok(());
+                }
                 let input = state.prompt.state.value().to_string();
                 state.prompt.reset(term)?;
 
-                // 1. Stash index (stash@0: index, ...)
-                let mut cmd = Command::new("git");
-                cmd.args(["stash", "push", "--staged"]);
-                state.run_cmd(term, &[], cmd)?;
+                let need_to_stash_index = is_something_staged(&state.repo)?;
 
-                // 2. Stash everything else (stash@0: worktree, stash@1: index, ...)
+                if need_to_stash_index {
+                    let mut cmd = Command::new("git");
+                    cmd.args(["stash", "push", "--staged"]);
+                    state.run_cmd(term, &[], cmd)?;
+                }
+
                 let mut cmd = Command::new("git");
                 cmd.args(["stash", "push", "--include-untracked"]);
                 if !input.is_empty() {
@@ -49,10 +56,11 @@ impl OpTrait for StashWorktree {
                 }
                 state.run_cmd(term, &[], cmd)?;
 
-                // 3. Pop stash with index (at stash@1)
-                let mut cmd = Command::new("git");
-                cmd.args(["stash", "pop", "1"]);
-                state.run_cmd(term, &[], cmd)?;
+                if need_to_stash_index {
+                    let mut cmd = Command::new("git");
+                    cmd.args(["stash", "pop", "1"]);
+                    state.run_cmd(term, &[], cmd)?;
+                }
             }
             Ok(())
         };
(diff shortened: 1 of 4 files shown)


altsem:
Trying to do some digging if there's a better way to handle all this.
There's multiple places where things might fail. What if there's an ongoing merge etc?

If only git had a `git stash push --worktree` :man_facepalming:.
The `git stash pop 1` is scary because it assumes nothing went wrong.

One idea is to return an error code from run_cmd and react to it.

I found this function in the git source code:
https://github.com/git/git/blob/d6fd04375f9196f8b203d442f235bd96a1a068cc/builtin/stash.c#L1288

altsem:
The C function does something like this I believe (123 would be the pid):
```bash
$ git diff --name-only | GIT_INDEX_FILE=.git/index.stash.123 git update-index --ignore-skip-worktree-entries --add --remove --stdin
```
And then it writes a "stash commit" somehow after.
Although doing this results in not just the diffed file being stage.

```bash
$ GIT_INDEX_FILE=.git/index.stash.123 git status
On branch fix/stash-working-tree-edge-cases
Your branch is up to date with 'golden-expiriensu/fix/stash-working-tree-edge-cases'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	deleted:    .cliffignore
	deleted:    .envrc
	deleted:    .github/dependabot.yml
	deleted:    .github/workflows/cachix.yml
...
```

Git creates this `.git/index.stash.123` file temporarily.

altsem:
Maybe just changing the `state.run_cmd()` function to actually return an error if the command failed is the way to go. :thinking: 
That way we'd just break execution in the `stash_worktree` fn.

Would not have to modify anything of what you already wrote in this case I think.

src/ops/stash.rs
```diff
@@ -33,25 +34,31 @@ impl OpTrait for StashWorktree {
     fn get_action(&self, _target: Option<&TargetData>) -> Option<Action> {
         let update_fn = move |state: &mut State, term: &mut Term| -> Res<()> {
             if state.prompt.state.status().is_done() {
+                if is_working_tree_empty(&state.repo)? {
```
altsem, line 37: There's no need to initiate a prompt when `is_working_tree_empty`.
See `state.prompt.set` below at line 70

altsem (changes requested):

src/ops/stash.rs
golden-expiriensu: Yeah, that makes sense. Do we want to show something to user? Like "Cannot stash: working tree is empty"

altsem: Sounds like a good idea!

src/tests/mod.rs
```diff
@@ -487,21 +487,44 @@ mod stash {
...
-                    key('i'),
-                    key_code(KeyCode::Enter),
-                    key('z'),
-                    key('p'),
-                    key('1'),
+                    key('w'),
+                    key('t'),
+                    key('e'),
+                    key('s'),
+                    key('t'),
```
altsem, line 527: I'd love to solve the syntax for sending these events somehow.
Would need one for keybind config anyway.
Something like, "jjjuzwtest<enter>".
And perhaps a test helper fn to select lines:
`select("Unstaged changes.*");`

Doesn't belong in this PR though.

altsem (approved):

src/ops/stash.rs
golden-expiriensu: I have encountered very strange situation.., can you try to run current tests and after that test things locally? Locally I am getting expected error message when working tree is empty, but there is no message in test. Did I miss something?

src/tests/mod.rs
golden-expiriensu: Wow, this sounds awesome! Implementing "jjjuzwtest<enter>" doesn't sound very hard tho. Selecting on the other hand looks pretty hard in my head, since we have everything inside the terminal. Does it have something like `read_to_string`? I need to research it a little bit

altsem: It's all in an in-memory buffer for tests. Seems a bit hard to work with.

src/ops/stash.rs
altsem: The test tries to `zw` and then input `test` and Enter.
It breaks after the change with the prompt I noticed. Pressing keys while the error is up will close it :) Line 500

src/ops/stash.rs
```diff
@@ -36,39 +41,88 @@ impl OpTrait for StashWorktree {
...
-            move |state: &mut State, _term: &mut Term| -> Res<()> {
-                state.prompt.set(PromptData {
+            move |state: &mut State, term: &mut Term| -> Res<()> {
+                if is_working_tree_empty(&state.repo)? {
+                    state
+                        .current_cmd_log_entries
+                        .push(Arc::new(RwLock::new(CmdLogEntry::Error(
+                            "Cannot stash: working tree is empty".to_string(),
+                        ))));
+                    state.prompt.reset(term)?;
```
altsem, line 76:
```suggestion
```
Should work without! There's no prompt open here yet.

altsem (changes requested):

src/ops/stash.rs
```diff
@@ -36,39 +41,88 @@ impl OpTrait for StashWorktree {
...
+                    return Ok(());
+                }
+                Ok(state.prompt.set(PromptData {
                     prompt_text: "Name of the stash:".into(),
                     update_fn: Rc::new(update_fn),
-                });
-                Ok(())
+                }))
```
altsem, line 82: Clippy is complaining about this in the build i noticed.
The `ci.sh` script has some goodies i run locally myself.

src/ops/stash.rs
golden-expiriensu: My bad 🤦‍♂️ Thanks for taking a look!

src/ops/stash.rs
golden-expiriensu: Indeed, done

src/ops/stash.rs
golden-expiriensu: Yeah, I was running only `cargo test`, now I will run ci script. Very strange rule to be honest, this should be delegated to compiler :)
Done

altsem (approved):

altsem:
Looked good! :)

Issue #97, opened 2024-03-26
Async git push, pull, fetch, etc


TheLostLambda:
At some point, it might be nice to make the calls to `git push`, `git pull`, and `git fetch` (anything that relies on the network) asynchronous! I'm stuck on a quite slow connection, so when I go to push my changes, the UI freezes up for a while!

I could just wait to do all of my pushing at once (after staging and committing everything outstanding), but it might be a nice quality of life feature to decouple the UI responsiveness from network access!

altsem:
This sounds reasonable! I realized things may become complicated, as you don't want to make the cursor move unexpectedly after e.g. a Pull.

I guess Fetch and Push are easier?

TheLostLambda:
I agree there would have to be some thinking about where to place the cursor post-pull, but how is that currently done with reloading after pressing `g`?

I might still find it less jarring to have the cursor / view change post pull (it does when I press `g` already) than to have it hang during a pull, but perhaps I'm just not 100% understanding what you mean (I'm sure there is more nuance on the code-side)!

altsem:
There's two aspects I think.
1. the cursor might move unexpectedly on a refresh
2. A refresh might happen unexpectedly

Point 1. already happens. Gitu is pretty dumb about this, but it seems fine due to how 2. never happens. A refresh only happens after a command is executed, or a g is pressed.

For example, if a Pull happened async:
1. I press 'Fp'.
2. The pop up shows at the bottom:
`$ git pull`

Currently:
3. Everything is blocked until the command finishes.
4. The view is refreshed
5. The popup will close as a press a key

3. The command is running in the background
4. I might decide I want to delete an untracked file, so I move the cursor to it
5. The Pull finishes and a refresh happens, yanking the cursor to something I did not intend to delete.
6. Before I can react, I press `K`

altsem:
@TheLostLambda I implemented this (I think) :)
I'll leave it on this PR for a while in case you or anyone like to test it out, or have any input.
#108

TheLostLambda:
Exciting! I'll take a look and kick the tires!

PR #108, opened 2024-03-31
Feat/non blocking

altsem:
closes: #97

Now fetch/pull/push can run async. Still only one git command (non-reads) can be run at a time. The popup will linger as long as its running, and show on-top of any menu too.

Running fetch as well as viewing the log menu:
image

Attempting to run a second `git fetch --all`:
image

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,13 +14,12 @@ pub mod term;
 mod tests;
 mod ui;
 
-use crossterm::event::{self};
+use crossterm::event::{self, Event};
 use git2::Repository;
 use items::Item;
 use itertools::Itertools;
-use ops::{Action, Op};
-use state::State;
-use std::{borrow::Cow, error::Error, iter, path::PathBuf, process::Command, rc::Rc};
+use ops::Action;
+use std::{borrow::Cow, error::Error, iter, path::PathBuf, process::Command, time::Duration};
 use term::Term;
 
 //                                An overview of Gitu's ui and terminology:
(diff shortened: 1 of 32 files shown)


TheLostLambda:
This is looking pretty good! I slammed my ping horribly high with a fun command I just discovered: `sudo tc qdisc add dev wlan0 root netem delay 500ms` (https://www.baeldung.com/linux/network-failures-simulation) and it behaved as expected, with the one "running" message at the bottom, and it added that error message when I tried to run another command whilst that one was in progress!

There are some crazy things that would maybe work, like trying to stage a file whilst waiting for a fetch, but I don't know if I want to discover weird race conditions in git right now... I think this is pretty good with nothing freezing on those network commands anymore!

Even if I can't stage files or patches whilst waiting on a network command, I now get a nice message why (a command is already running) and the UI stays responsive! Great (and quick) work!

TheLostLambda:
From the code perspective, I think it's probably not a real issue, but if you want to avoid that polling every 100ms, you might be able to use something like this: https://crates.io/crates/async-process

I've not actually used it myself before, but it looks super popular on crates.io and will use Rust's async await that I think in most cases avoids polling and sets up event listeners in the background!

If that looks like a mess to integrate though, then I think this is more than fine as is!

TheLostLambda:
(Looking closer, I think you'd need to also switch to listening to those crossterm events from an async `EventStream`, then `join` the event-stream and pending command futures — essentially waiting for whichever of the two comes first. But it looks like using that `EventStream` also means pulling in something like `async-std` or `tokio`, and that might be more work than is worth it!)

altsem:
Git does keep some sort of lock-file I discovered. Perhaps its fine. Would need to keep track of multiple running processes. Another caveat I found now, is that you can't open a hunk with a text editor while it is e.g. fetching.

I had a go at turning things async before. Listening to those crossterm events async and such. I think my PoC turned garbage, it's probably doable and nice. Might try it again in the future :+1:

I'll go ahead and merge it and close the issue. Ty for the help testing and researching! :)

Issue #114, opened 2024-04-05
Discarding staged new file is not working


golden-expiriensu:
Reproduction:
```
 danil   fix/stash-worktree  ~/offw/rust/gitu  echo "meowmeow" > cat.txt
 danil   fix/stash-worktree ?:1  ~/offw/rust/gitu  gitu -k "jsjKy<enter>"
```
Yields error:
![Screenshot from 2024-04-05 13-32-21](https://github.com/altsem/gitu/assets/83108115/ec7f6612-86f3-4951-a1c9-0a92108b5a4e)

I think we want to `git restore --staged <file>` followed by `std::fs::remove_file(file)`, right?

PR #115, opened 2024-04-05
fix: if file is new then unstage it before discarding, use clean command instead of fs for removing untracked file

golden-expiriensu:
Fixes: https://github.com/altsem/gitu/issues/114

### DIFF
--- a/src/ops/discard.rs
+++ b/src/ops/discard.rs
@@ -1,7 +1,7 @@
-use super::{cmd, cmd_arg, Action, OpTrait};
+use super::{cmd, cmd_arg, unstage::unstage_file_cmd, Action, OpTrait};
 use crate::{items::TargetData, state::State, term::Term};
 use derive_more::Display;
-use std::{ffi::OsStr, path::PathBuf, process::Command, rc::Rc};
+use std::{ffi::OsStr, process::Command, rc::Rc};
 
 #[derive(Default, Clone, Copy, PartialEq, Eq, Debug, Display)]
 #[display(fmt = "Discard")]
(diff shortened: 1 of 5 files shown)


altsem:
Nice! I'll have a look :)

altsem (commented):
Good to see some love for these. I did some testing and seems to work!
Is `git rm --force *file*` a shorter way to achieve the same thing?

src/ops/discard.rs
```diff
@@ -10,20 +10,22 @@ impl OpTrait for Discard {
...
+                    git2::Delta::Added => Rc::new(move |state: &mut State, term: &mut Term| {
+                        let file = d.new_file.as_os_str();
+                        state.run_cmd(term, &[], unstage_file_cmd(file))?;
+                        state.run_cmd(term, &[], clean_file_cmd(file))
+                    }),
+                    _ => {
+                        if d.old_file == d.new_file {
+                            cmd_arg(checkout_file_cmd, d.old_file.into())
+                        } else {
+                            // TODO Discard file move
```
altsem, line 25: Optional: I think this old_file == new_file check and TODO can just be removed. I can't recall it ever happening, and should now be covered by your match on line 15 anyway!

src/ops/discard.rs
```diff
@@ -10,20 +10,22 @@ impl OpTrait for Discard {
...
             Some(TargetData::Branch(r)) => cmd_arg(discard_branch, r.into()),
-            Some(TargetData::File(f)) => Rc::new(move |state: &mut State, _term: &mut Term| {
-                let path = PathBuf::from_iter([
-                    state.repo.workdir().expect("No workdir").to_path_buf(),
-                    f.clone(),
-                ]);
-                std::fs::remove_file(path)?;
-                state.screen_mut().update()
-            }),
+            Some(TargetData::File(f)) => cmd_arg(clean_file_cmd, f.into()),
```
altsem, line 13: Optional: I had a look at why there's TargetData::File and TargetData::Delta, and found some refactoring opportunity. It should be possible to remove TargetData::File completely removed from the code-base!

golden-expiriensu: There is one place where TargetData::File is created and it is in items_list. I tried to replace it with delta but got confused pretty quick. I think I need to look into that one more time later...

golden-expiriensu:
> Good to see some love for these. I did some testing and seems to work! Is `git rm --force *file*` a shorter way to achieve the same thing? i.e.:
> 
> * added untracked (git clean --force _file_)
> * added unstaged (git rm --force _file_)
> * added staged (git rm --force _file_)

This is a beautiful solution, thank you!

src/ops/discard.rs
golden-expiriensu: Done! Is it intentional that it shows moved file as a pair of added-deleted?

altsem: I think there must be some configuration to do move detection with libgit2.
It doesn't seem to happen out of default. Just a matter of implementing/fixing

altsem (approved):
I'll squash these with a more user-friendly message as the "fix: " commit is the one that'll appear in the changelog.
:rocket:

PR #127, opened 2024-04-11
Feat/stash args

altsem:
@golden-expiriensu tagging you here in case you want to review :)

The goal here was to basically add a `cmd.args(args);` to relevant stash commands.
This will include args enabled in the menu:
```
Arguments
-a Also save untracked and ignored files (--all)
-u Also save untracked files (--include-untracked)
```

`--include-untracked` is enabled per default (and can now be toggled off).

I tried to simplify the code regarding the prompt to more align with updates in other modules!

### DIFF
--- a/src/default_config.toml
+++ b/src/default_config.toml
@@ -131,6 +131,8 @@ reset_menu.reset_hard = ["h"]
 reset_menu.quit = ["q", "<esc>"]
 
 root.stash_menu = ["z"]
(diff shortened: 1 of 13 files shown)


golden-expiriensu:
Yeah, this looks much more consistent now! Really cool that args are not hardcoded anymore

golden-expiriensu (approved):

PR #130, opened 2024-04-12
feat: copy commit hash with "y", move Show Refs to "Y"

### LINKED ISSUE #119: Copy commit hash to clipboard
It'd be useful to be able to copy a commit hash.
Select anything with a commit hash, press `y` and it would be copied to your clipboard.

`y` is already bound to `show_refs` in the current config, but I suggest moving `show_refs` to maybe `Y`.

FrancoRav:
Add "copy commit hash" feature, bound to "y" by default. Changes default keybinding for show refs to "Y", as suggested in #119 
Closes #119


altsem (changes requested):
Looks great! Just some minor things.

`cargo fmt --check` seems to be failing, as well as `cargo deny check`. (these are found in the `ci.sh` script).

`BSL-1.0` could be added to the allowed list of licences in [deny.toml](deny.toml)

Perhaps the added changelog (`git cliff --unreleased`) could be a bit more descriptive:
```
### 🚀 Features

- Add copy commit hash feature
```
Something like:
Copy commit hash with "y", move Show refs to "Y"
?

..just to convey that the binding was moved as part of the changelog!

FrancoRav:
> Looks great! Just some minor things.
> 
> `cargo fmt --check` seems to be failing, as well as `cargo deny check`. (these are found in the `ci.sh` script).
> 
> `BSL-1.0` could be added to the allowed list of licenses in [deny.toml](deny.toml)
> 
> Perhaps the added changelog (`git cliff --unreleased`) could be a bit more descriptive:
> 
> ```
> ### 🚀 Features
> 
> - Add copy commit hash feature
> ```
> 
> Something like: Copy commit hash with "y", move Show refs to "Y" ?
> 
> ..just to convey that the binding was moved as part of the changelog!

I had never used git cliff before. Is there any way to change the changelog message without force pushing to change the commit message?

altsem:
Hmm perhaps, it'll be finicky though. The release pipeline I've built is all based on these generated messages, there's ways to make exceptions to it. I'm not all too familiar.
But otherwise force-pushing and changing the message is a pretty fix IMO.

FrancoRav:
I force pushed to reword the commits for the Changelog.
Also, I just noticed the `Clipboard::new()` function fails in the test environment as it can't access any clipboard (the error is `Unknown error while interacting with the clipboard: $DISPLAY variable not set and no value was provided explicitly`). Not sure how to solve this
I was thinking making the Clipboard an `Option<Clipboard>` and leaving None if it errors, but then the feature could not be tested.
Using Option for that is probably necessary anyway, since we need error handling in case it happens in real use, but it still doesn't fix the problem of how to test the feature. Any ideas?

altsem:
There's not a lot of code, and if we're to test anything, I guess it could be the message?
Seems fine wrapping it in an Option!

FrancoRav:
> There's not a lot of code, and if we're to test anything, I guess it could be the message? Seems fine wrapping it in an Option!

That's done now. It should pass ci now, and it should be working correctly.
I'm still not sure about the way I am displaying the info and erros in the CmdLog (the methods I added to the State), so if you suggest any changes I can look into that

src/ops/copy_hash.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+fn copy_hash(r: String) -> Option<Action> {
+    Some(Rc::new(move |state, _term| {
+        match &mut state.clipboard {
+            Some(cb) => {
+                cb.set_text(r.clone())?;
+                state.display_info("Commit hash copied to clipboard".to_owned());
+            }
+            None => state.display_error("Clipboard not available".to_owned()),
```
altsem, line 28: I think this would be handled by just returning Err with the message too.
Seems fine either way to me. :+1:

altsem (approved):

Issue #125, opened 2024-04-11
Improve support for tags


frank-lenormand:
Hi,

The log view shows tags only if they happen to point at a revision in view.

The revision view does not show that it is pointed to by a tag.

I didn’t see any obvious way of displaying all the tags in the repository either, maybe I missed it.

It’d be convenient to put more focus on tags:

- display tags in the reference view (they are stored in `.git/refs/tags` after all)
- have a dedicated tag view that shows all the information related to it (signing, annotation…)
- allow jumping to a tag view from the log/revision view
- showing that the current revision in the revision view is pointed to by a tag

golden-expiriensu:
@altsem I think it's worth adding tags to show_refs right?

Also maybe segregate branches on local and remotes like in magit?:
![image](https://github.com/altsem/gitu/assets/83108115/b5a54454-8e9a-4aed-92b3-9f714aad4138)

The issue for me is that in my company we name branches like "feature/JIRA-XXX-quite-long-description-of-the-feature" and pretty often two branches (local and remote) don't fit onto one line
<img width="1710" alt="Screenshot 2567-04-12 at 13 20 38" src="https://github.com/altsem/gitu/assets/83108115/1e181efc-d645-469d-a287-fa4761cd3939">
So having them in two different sections would make it more neat

PR #134, opened 2024-04-14
feat: 3 sections in show refs screen: branches, remotes, tags

golden-expiriensu:
Resolves: https://github.com/altsem/gitu/issues/125

### DIFF
--- a/src/screen/show_refs.rs
+++ b/src/screen/show_refs.rs
@@ -62,7 +62,7 @@ fn create_remotes_sections<'a>(
     header_style: &'a StyleConfigEntry,
     item_style: &'a StyleConfigEntry,
 ) -> Res<impl Iterator<Item = Item> + 'a> {
-    let all_remotes = create_references_section(&repo, Reference::is_remote, item_style)?;
+    let all_remotes = create_references_section(repo, Reference::is_remote, item_style)?;
     let mut remotes = HashMap::new();
     for remote in all_remotes {
         let name = match remote.id.split_once('/') {
@@ -115,22 +115,29 @@ where
         .filter(filter)
         .map(move |reference| {
             let name = Span::styled(reference.shorthand().unwrap().to_string(), style);
-            let head = repo.head().ok();
-
-            let prefix = Span::raw(
-                if reference.target() == head.as_ref().and_then(Reference::target) {
-                    "* "
-                } else {
-                    "  "
-                },
-            );
 
             Item {
                 id: name.clone().content,
-                display: Line::from(vec![prefix, name.clone()]),
+                display: Line::from(vec![create_prefix(repo, &reference), name.clone()]),
                 depth: 1,
                 target_data: Some(TargetData::Branch(name.content.into())),
                 ..Default::default()
             }
         }))
+}
+
+fn create_prefix(repo: &Repository, reference: &Reference) -> Span<'static> {
+    let head = repo.head().ok();
+
+    Span::raw(if repo.head_detached().unwrap_or(false) {
+        if reference.target() == head.as_ref().and_then(Reference::target) {
+            "? "
+        } else {
+            "  "
+        }
+    } else if reference.name() == head.as_ref().and_then(Reference::name) {
+        "* "
+    } else {
+        "  "
+    })
 }
--- a/src/screen/show_refs.rs
+++ b/src/screen/show_refs.rs
@@ -1,4 +1,8 @@
-use std::{iter, rc::Rc};
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    iter,
+    rc::Rc,
+};
 
 use super::Screen;
 use crate::{
@@ -26,21 +30,14 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
                 depth: 0,
                 ..Default::default()
             })
-            .chain(create_references_section_items(
+            .chain(create_references_section(
                 &repo,
                 Reference::is_branch,
                 &style.branch,
             )?)
-            .chain(iter::once(Item {
-                id: "remote_branches".into(),
-                display: Line::styled("Remote".to_string(), &style.section_header),
-                section: true,
-                depth: 0,
-                ..Default::default()
-            }))
-            .chain(create_references_section_items(
+            .chain(create_remotes_sections(
                 &repo,
-                Reference::is_remote,
+                &style.section_header,
                 &style.remote,
             )?)
             .chain(iter::once(Item {
@@ -50,7 +47,7 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
                 depth: 0,
                 ..Default::default()
             }))
-            .chain(create_references_section_items(
+            .chain(create_references_section(
                 &repo,
                 Reference::is_tag,
                 &style.tag,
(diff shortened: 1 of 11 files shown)


golden-expiriensu:
@altsem PTAL, can I start writing some tests or you would like me to change something (not sure if current approach is the best one)

src/screen/show_refs.rs
```diff
@@ -18,50 +18,40 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
...
-                                Span::styled(name.to_string(), &style.remote)
-                            } else {
-                                Span::raw("")
-                            }
-                        } else {
-                            Span::raw("")
-                        };
+                    .filter(Reference::is_tag)
+                    .map(|tag| {
+                        let name = Span::styled(tag.name().unwrap().to_string(), &style.branch);
```
altsem, line 50:
```suggestion
                        let name = Span::styled(tag.name().unwrap().to_string(), &style.tag);
```

src/screen/show_refs.rs
```diff
@@ -18,50 +18,40 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
...
             })
+            .chain(branches(&repo, Some(BranchType::Local), &style.branch)?)
+            .chain(iter::once(Item {
+                id: "remote_branches".into(),
+                display: Line::styled("Remote".to_string(), &style.section_header),
+                section: true,
+                depth: 0,
+                ..Default::default()
+            }))
+            .chain(branches(&repo, Some(BranchType::Remote), &style.branch)?)
```
altsem, line 37:
```suggestion
            .chain(branches(&repo, Some(BranchType::Remote), &style.remote)?)
```
There's a style.remote.
In the log remotes / tags / branches are colored after branch/remote/tag configs.
Should this page use the same?

src/screen/show_refs.rs
```diff
@@ -18,50 +18,40 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
...
+                id: "local_branches".into(),
                 display: Line::styled("Branches".to_string(), &style.section_header),
                 section: true,
                 depth: 0,
                 ..Default::default()
             })
+            .chain(branches(&repo, Some(BranchType::Local), &style.branch)?)
+            .chain(iter::once(Item {
+                id: "remote_branches".into(),
+                display: Line::styled("Remote".to_string(), &style.section_header),
```
altsem, line 32: In magit, there's one section per remote
```
Remote FrancoRav (git@github.com:FrancoRav/gitu.git) (4)…
Remote golden-expiriensu (git@github.com:golden-expiriensu/gitu.git) (6)…
Remote jondysinger (git@github.com:jondysinger/gitu.git) (2)
  feat/confirm-quit               test: dont test quit confirmation for now (off by default)…
  master                          ci: set 'release' environment for release workflow…

```

It doesn't need to look the same (maybe in the future?), but would it make more sense to label this "Remote branches" for now?

altsem (commented):
I think it looks fine :)
Maybe there's a nice way to sort the tags.

Added some ideas

src/screen/show_refs.rs
golden-expiriensu: Yeah, it would be very cool to have them separated, lemme try

src/screen/show_refs.rs
golden-expiriensu: Sure, I totally missed it

src/screen/show_refs.rs
golden-expiriensu: I am on it

golden-expiriensu:
This is how it looks when every section uses its dedicated style, what do you think?

Screenshot 2567-04-14 at 18 47 48

altsem:
It is a first step at least :+1:
One thing, I noticed that both the remote and the local branch are prefixed with a `*`.
If I checkout a remote branch, a local branch would still have a `*`, even though the HEAD is detached.

src/screen/show_refs.rs
golden-expiriensu: Done

src/screen/show_refs.rs
golden-expiriensu: Done

src/screen/show_refs.rs
golden-expiriensu: Done

golden-expiriensu:
Okay, this should be much better now:
Remotes section is now segregated by the prefix before first slash (origin/main -> origin, text-text/main -> text-text) where header is capitalized prefix (origin -> Origin, text-text -> Text-text).
When head is detached (.git/HEAD contains commit hash instead of ref name) there is two options:
1. Show some special symbol on every match by target (current solution with "? " prefix)
2. Compare reference to last target in reflog?

src/screen/show_refs.rs
```diff
@@ -18,57 +22,122 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
...
+
+        match remotes.entry(name.to_string()) {
+            Entry::Vacant(entry) => {
+                entry.insert(vec![remote]);
+            }
+            Entry::Occupied(mut entry) => {
+                entry.get_mut().push(remote);
+            }
+        }
+    }
```
altsem, line 81: There's an issue (since it's a hashmap). The order of the items will be randomized on a refresh.

These functions exist in git2, might be a nicer way than splitting by `/`:
```rust
repo.remotes()?.iter().map(|remote_name: Option<&str>| {
    // Why does find_remote accept a &str, when remote_name might be None?
    let remote = repo.find_remote(remote_name)?;
    todo!();
});

repo.remotes()?.iter_bytes().map(|remote_name: &[u8]| {
    // The bytes variant is non-optional, but how to get the data on a remote from here?
});
```

The git2 API is a bit messy sometimes. I'd like to start experimenting with Gitoxide some time soon.

src/screen/show_refs.rs
```diff
@@ -18,57 +22,122 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
...
+                Reference::is_branch,
+                &style.branch,
+            )?)
+            .chain(create_remotes_sections(
+                &repo,
+                &style.section_header,
+                &style.remote,
+            )?)
+            .chain(iter::once(Item {
+                id: "tags".into(),
```
altsem, line 44: On the status page there's spacing between sections. This is just a `items::blank_line()` inserted between them.

src/screen/show_refs.rs
```diff
@@ -18,57 +22,122 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
...
+                entry.get_mut().push(remote);
+            }
+        }
+    }
+
+    Ok(remotes.into_iter().flat_map(move |(name, items)| {
+        let header = capitalize(&name);
+        iter::once(Item {
+            id: name.into(),
+            display: Line::styled(header, header_style),
```
altsem, line 87: I'd just copy Magit and have these headers be e.g. `Remote origin`, makes it clear that it is a remote.
It is simpler to not capitalize the origin names I think.
Tested it, and it is actually possible to have `Origin` and `origin` simultaneously in git.

src/screen/show_refs.rs
```diff
@@ -18,57 +22,122 @@ pub(crate) fn create(config: Rc<Config>, repo: Rc<Repository>, size: Rect) -> Re
...
+            }
+        }))
+}
+
+fn create_prefix(repo: &Repository, reference: &Reference) -> Span<'static> {
+    let head = repo.head().ok();
+
+    Span::raw(if repo.head_detached().unwrap_or(false) {
+        if reference.target() == head.as_ref().and_then(Reference::target) {
+            "? "
```
altsem, line 134: I like this! More useful than nothing. :+1: 
In magit it'd just say:
image

altsem (changes requested):
I think I found some points, hope they make sense.

src/screen/show_refs.rs
golden-expiriensu: Done, also don't show section if its empty (assuming here that there is always at least 1 local branch)

src/screen/show_refs.rs
golden-expiriensu: Done

src/screen/show_refs.rs
golden-expiriensu: I glad you liked it :)

src/screen/show_refs.rs
golden-expiriensu: Done

altsem:
I had a Quick Look at the code now. Will do some testing after work maybe 👍

altsem (approved):

Issue #162, opened 2024-05-03
[Bug] incorrect diff if file uses CRLF line endings


rzikm:
On Windows

gitu - shows changes on every single line:
![image](https://github.com/altsem/gitu/assets/32671551/75c9f8f3-5bca-4050-a522-7a90aab86e0d)

git:
![image](https://github.com/altsem/gitu/assets/32671551/9b1638fd-6b68-4e5f-a74e-dadbed83da81)

altsem:
From what I could conclude occurs when the line-ending was changed from/to crlf/lf.
I suppose here the `\r` is highlighted as removed (though it's invisible).

```rust
#[test]
fn crlf_diff() {
    let ctx = TestContext::setup_init();
    commit(ctx.dir.path(), "clrf.txt", "unchanged\r\nunchanged\r\n");
    fs::write(ctx.dir.child("clrf.txt"), "unchanged\nunchanged\n").unwrap();

    snapshot!(ctx, "jj<tab>");
}
```

bshashank:
I too tried gitu recently and hit this bug. One thing to mention is that you can only reproduce the issue when autocrlf is turned on:
```
$ git config --local core.autocrlf
true
```
Or you have setup `.gitattributes` to do the equivalent. ([more context](https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/))

That's when the git diff is much smaller than `gitu`. Since git will convert all the `\r\n` to `\n` when writing back to object directory.

I think the only way to solve for this to read these configs from git (via local/global configs and the `.gitattributes` file) and ignore `\r` differences if `autoclrf` is true or input.

I saw elsewhere that we want to use gitoxide as the git layer, [which provides this conversion](https://github.com/Byron/gitoxide/blob/048e43e26908b0572852a75780a451460dc152ff/gix-diff/src/blob/platform.rs#L114). But that change would also probably require changing from the similar diff library to using [imara-diff](https://docs.rs/imara-diff/latest/imara_diff/)

altsem:
That's great input @bshashank, ty!

I've been thinking of pulling in Imara-diff instead and/or using Gitoxide.
It should be easier now ever since a large chunk of the inline diffing was re-written to no rely so much on Similar.
However, there might be an easy fix before moving there.

PR #169, opened 2024-05-08
fix: bad diffs when git's `autocrlf` was enabled

altsem:
Before all CRLF would already be converted to LF after the diff was made (due to line 85 / how Ratatui's Text works).
Now this happens before the diff is made.

resolves: #162

I might've overlooked something, but I think this would work.
@bshashank, @rzikm

### DIFF
--- a/src/git/diff.rs
+++ b/src/git/diff.rs
@@ -154,12 +154,13 @@ fn diff_files(
     config: &Config,
     delta: &Delta,
 ) -> Res<Vec<Rc<Hunk>>> {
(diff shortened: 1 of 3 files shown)


rzikm:
This fixes the problem in my use case. Thanks!

bshashank (approved):
I tested it on my windows box and looks correct now even with `autocrlf=true`.  Only lines added by me are shown (by both gitu and git)

Thanks for the quick turnaround!

PR #218, opened 2024-07-12
feat(config): collapsed sections

stackmystack:
I stringed this in a couple of minutes, so I am not sure whether this is enough.

I tried to look for tests on collapsed sections, and couldn't find something useful.

I'll be happy to add a test if you have a good idea on what to do.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -22,6 +22,7 @@ pub(crate) struct Config {
 pub struct GeneralConfig {
     pub always_show_help: BoolConfigEntry,
     pub confirm_quit: BoolConfigEntry,
+    pub collapsed_sections: Vec<String>,
 }
 
 #[derive(Default, Debug, Deserialize)]
--- a/src/default_config.toml
+++ b/src/default_config.toml
@@ -5,6 +5,9 @@
 [general]
 always_show_help.enabled = false
 confirm_quit.enabled = false
+# To collapse untracked files by default. e.g.:
+#   collapsed_sections = ["untracked", "Recent commits", "branch_status"]
+collapsed_sections = []
 
 [style]
 # fg / bg can be either of:
@@ -29,7 +32,7 @@ diff_highlight.tag_new = { fg = "green", mods = "BOLD" }
 diff_highlight.unchanged_old = { mods = "DIM" }
 diff_highlight.unchanged_new = { mods = "DIM" }
 diff_highlight.changed_old = { fg = "red" }
-diff_highlight.changed_new = { fg = "green"}
+diff_highlight.changed_new = { fg = "green" }
 
 syntax_highlight.enabled = true
 syntax_highlight.attribute = { fg = "yellow" }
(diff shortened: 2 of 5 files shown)


altsem:
Makes sense, I'll test this out soon!
Maybe a more descriptive name in the config file would be nice?

```
[general]
always_show_help.enabled = false
always_show_help.enabled = false
confirm_quit.enabled = false
confirm_quit.enabled = false
collapsed = []
```

stackmystack:
> Makes sense, I'll test this out soon! Maybe a more descriptive name in the config file would be nice?

`collapsed_sections = []` ? `collapsed_on_startup = []` ?

> ```
> [general]
> always_show_help.enabled = false
> always_show_help.enabled = false
> confirm_quit.enabled = false
> confirm_quit.enabled = false
> collapsed = []
> ```

I didn't quite get your comment here, you duplicated some keys and kept `collapsed`.

altsem:
What the hell did I copy paste x)

Yes, just changing the name of it was my idea. Perhaps 'collapsed_sections' is good enough :)

stackmystack:
Done.

altsem:
An issue I've thought of is that there's no real scheme for ids of items in gitu.
Some are just a copy of the text they contain, some are defined as snake case (like "untracked").

These were put in just in order to have a (hopefully) unique reference to an item that would remain between reloads of data. (so that collapsed sections will remain collapsed on a reload).

Perhaps only these defined sections on the status screen should be configurable?
Section in gitu could be a delta, a hunk etc.

Currently the sections seen on the status screen are also not consistent:
```rust
#[test]
fn collapsed_sections_config() {
    let mut ctx = TestContext::setup_clone();
    ctx.config().general.collapsed_sections = vec![
        // TODO Are we doing snake_case or not?
        "untracked".into(),
        "Recent commits".into(),
        "branch_status".into(),
        // TODO rebase / revert/ merge conlict?
    ];
    fs::write(ctx.dir.child("untracked_file.txt"), "").unwrap();

    snapshot!(ctx, "");
}
```

It would be nice if it's documented in the default config too, exactly which ones can configured as collapsed.

Does that make sense?

stackmystack:
> An issue I've thought of is that there's no real scheme for ids of items in gitu. Some are just a copy of the text they contain, some are defined as snake case (like "untracked").

Yes that's why I thought. I think IDs should be snake case, but for the config part, it shouldn't matter. We can take the user text and make it `snake_case`.


> Perhaps only these defined sections on the status screen should be configurable? Section in gitu could be a delta, a hunk etc.

Yeah, well, we can never know what users want, and if we go through the route of unique snake_case IDs, then theoretically it would even be less specific code since we don't have to manage exceptions?

> It would be nice if it's documented in the default config too, exactly which ones can configured as collapsed.

I added in the docs the one you included in the tests since they definitely make the most sense ATM.

And since we're at collapsed sections, I noticed that when I press `g` to refresh the screen and get modifications, the delta for files are open by default, and I think it should be closed by default because in most scenarios, this is what you want.

altsem (changes requested):

altsem:
Seems fine to start with these.
The deltas showing per default is a bug from before I think? Could be done in another issue and PR.

altsem:
superseded by #276

PR #230, opened 2024-07-20
test(all): restore terminal state when finished.

stackmystack:
Runnin `cargo t` used to leave the terminal in a bad state: no mouse input is interpreted right, scrolling is gone, etc.


altsem (approved):
Nice! Looks like I forgot the run these in tests, they're done in main.rs only.

PR #231, opened 2024-07-23
fix: instant fixup commit would not work with some versions of Git

altsem:
@stackmystack FYI. It seems that when rebasing, skipping `--interactive` when doing `--autosquash` breaks in some versions of Git.
I'm reverting it back to use `GIT_SEQUENCE_EDITOR`!

### DIFF
--- a/src/ops/commit.rs
+++ b/src/ops/commit.rs
@@ -118,8 +118,15 @@ impl OpTrait for CommitInstantFixup {
 
 fn rebase_autosquash_cmd(rev: &OsStr) -> Command {
     let mut cmd = Command::new("git");
-    cmd.args(["rebase", "--autostash", "--keep-empty", "--autosquash"]);
+    cmd.args([
+        "rebase",
+        "-i",
+        "--autostash",
+        "--keep-empty",
+        "--autosquash",
+    ]);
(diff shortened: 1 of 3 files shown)


stackmystack:
What versions exactly? I mean of they're old enough maybe you shouldn't care that much? Gitu is a developer (human) tool and I'd say there should be limits to how backward compatible with git it needs to be.

altsem:
It broke in CI on GH Action's default ubuntu and Mac images, as well as on my Mac laptop with git 2.39.3 (Apple Git-146).
It did succeed on the Windows CI, as well as on my desktop running Git 2.45.2

From the man page of git rebase:

```
BEHAVIORAL DIFFERENCES
       git rebase has two primary backends: apply and merge. (The apply backend used to be known as the am backend, but the name led to confusion as it looks like a verb instead of a noun. Also, the merge
       backend used to be known as the interactive backend, but it is now used for non-interactive cases as well. Both were renamed based on lower-level functionality that underpinned each.) There are some
       subtle differences in how these two backends behave
```

Confusing as heck :D
But it appears to behave nicely now :)

stackmystack:
The famous consistency of git …

Issue #224, opened 2024-07-13
[Feature request] Support "renamed" status for a file that have been moved/renamed via "git mv"


meseck:
The current behavior is that the renamed file has both statuses "added" for the new name and "deleted" for the old name.

However, using "git status" will display the file with a "renamed" status, clearly indicating that the file was moved or renamed correctly while preserving its history. 

Thanks for this tool, it's awesome! 🚀

meseck:
Sorry that I accidentally opened multiple duplicates of this issue. I really don't know how it happened. 🤷 😅 
I hope it's easy to remove them.

altsem:
Np! This is something I've been wanting to get right. Think there's some "move detection" you can configure in libgit2 to make it work. Haven't looked into it too deep.

z0al:
Calling `find_similar` in `diff_staged` [here](https://github.com/altsem/gitu/blob/a30c2aea8873d1fbf025e80c7355049b56db0f37/src/git/mod.rs#L128) did the trick:

```rust
diff.find_similar(Some(&mut DiffFindOptions::new().renames(true)))?;
```

**Using `git status`**

<img width="663" alt="image" src="https://github.com/user-attachments/assets/0faf8df4-8249-471c-be55-34c47bde8f9e">

**Using `gitu`**

<img width="668" alt="image" src="https://github.com/user-attachments/assets/3775748d-0c1f-47b7-89c5-2b3bbe335ff2">

I tried to adjust the rendering to match the `git status` output e.g. `renamed old_file -> new_file` but my non-existent Rust experience failed me. Any pointers @altsem 😅?  also, the misalignment of the columns is super annoying.

altsem:
Cool! 🙌
That seems pretty easy to introduce, I should have a double-check and test it. I'm sure aligning the text can be done too.
I'll have a check in the coming days when I get the time :)

z0al:
Thanks @altsem. I will open a PR shortly
<img width="668" alt="image" src="https://github.com/user-attachments/assets/59a3e810-7f36-4645-a02f-1fbdb08345e2">

PR #234, opened 2024-07-29
feat(status): detect renamed files

z0al:
### Changes:
- Detect renamed files
- Align status columns
- Update test snapshots

Fixes https://github.com/altsem/gitu/issues/224

### Demo
image

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,7 +10,7 @@
 
   outputs = inputs@{ nixpkgs, flake-parts, ... }:
     flake-parts.lib.mkFlake { inherit inputs; } {
-      systems = [ "aarch64-linux" "x86_64-linux" "x86_64-darwin" ];
+      systems = [ "aarch64-linux" "x86_64-linux" "x86_64-darwin" "aarch64-darwin" ];
 
       perSystem = { config, pkgs, system, inputs', self', ... }:
         let
--- a/src/git/mod.rs
+++ b/src/git/mod.rs
@@ -1,4 +1,4 @@
-use git2::Repository;
+use git2::{DiffFindOptions, Repository};
 use itertools::Itertools;
 
 use self::{commit::Commit, diff::Diff, merge_status::MergeStatus, rebase_status::RebaseStatus};
@@ -121,10 +121,12 @@ pub(crate) fn diff_unstaged(config: &Config, repo: &Repository) -> Res<Diff> {
 pub(crate) fn diff_staged(config: &Config, repo: &Repository) -> Res<Diff> {
     let opts = &mut git2_opts::diff(repo)?;
 
-    let diff = match repo.head() {
+    let mut diff = match repo.head() {
         Ok(head) => repo.diff_tree_to_index(Some(&head.peel_to_tree()?), None, Some(opts))?,
         Err(_) => repo.diff_tree_to_index(None, None, Some(opts))?,
     };
+
+    diff.find_similar(Some(&mut DiffFindOptions::new().renames(true)))?;
 
     diff::convert_diff(config, repo, diff, false)
 }
--- a/src/ops/discard.rs
+++ b/src/ops/discard.rs
@@ -13,6 +13,7 @@ impl OpTrait for Discard {
             Some(TargetData::File(file)) => clean_file(file),
             Some(TargetData::Delta(d)) => match d.status {
                 git2::Delta::Added => remove_file(d.new_file),
+                git2::Delta::Renamed => rename_file(d.new_file, d.old_file),
                 _ => checkout_file(d.old_file),
             },
             Some(TargetData::Hunk(h)) => discard_unstaged_patch(h),
@@ -43,6 +44,18 @@ fn clean_file(file: PathBuf) -> Action {
         let mut cmd = Command::new("git");
         cmd.args(["clean", "--force"]);
         cmd.arg(&file);
+
+        state.close_menu();
+        state.run_cmd(term, &[], cmd)
+    })
+}
+
+fn rename_file(src: PathBuf, dest: PathBuf) -> Action {
+    Rc::new(move |state, term| {
+        let mut cmd = Command::new("git");
+        cmd.args(["mv", "--force"]);
+        cmd.arg(&src);
+        cmd.arg(&dest);
 
         state.close_menu();
         state.run_cmd(term, &[], cmd)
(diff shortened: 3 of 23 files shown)


src/tests/snapshots/gitu__tests__discard__discard_file_move.snap
```diff
@@ -19,7 +19,7 @@ expression: ctx.redact_buffer()
                                                                                 |
                                                                                 |
                                                                                 |
-                                                                                |
 ────────────────────────────────────────────────────────────────────────────────|
-$ git checkout HEAD -- new-file                                                 |
-styles_hash: 19cf9a9796ec3951
+$ git rm --force moved-file                                                     |
+rm 'moved-file'                                                                 |
+styles_hash: bc84efbf82da3787
```
z0al, line 25: I'm not sure what this is supposed to mean and if it's expected. Any idea @altsem ?

flake.nix
```diff
@@ -10,7 +10,7 @@
 
   outputs = inputs@{ nixpkgs, flake-parts, ... }:
     flake-parts.lib.mkFlake { inherit inputs; } {
-      systems = [ "aarch64-linux" "x86_64-linux" "x86_64-darwin" ];
+      systems = [ "aarch64-linux" "x86_64-linux" "x86_64-darwin" "aarch64-darwin" ];
```
z0al, line 13: This is to enable using `nix` on Apple Silicon

src/git/mod.rs
```diff
@@ -121,11 +121,13 @@ pub(crate) fn diff_unstaged(config: &Config, repo: &Repository) -> Res<Diff> {
 pub(crate) fn diff_staged(config: &Config, repo: &Repository) -> Res<Diff> {
     let opts = &mut git2_opts::diff(repo)?;
 
-    let diff = match repo.head() {
+    let mut diff = match repo.head() {
         Ok(head) => repo.diff_tree_to_index(Some(&head.peel_to_tree()?), None, Some(opts))?,
         Err(_) => repo.diff_tree_to_index(None, None, Some(opts))?,
     };
 
+    diff.find_similar(Some(&mut DiffFindOptions::new().renames(true)))?;
```
z0al, line 129: Another available option here is [copies detection](https://docs.rs/git2/latest/git2/struct.DiffFindOptions.html#method.copies). I left it out because I have never seen how that works in practice (in git cli) and I'm only interested in renames for now.

src/tests/snapshots/gitu__tests__discard__discard_file_move.snap
altsem: The "discard" module runs different git commands depending on context.
https://github.com/altsem/gitu/blob/86603c81fd015fc16d25c76757b99a87522a34fc/src/ops/discard.rs#L14-L17

It just doesn't properly handle moved files yet.
I tested this out in Magit now to see how it handles it, it seems:

Unstaged:
image

Staged:
image

And discarding:
image

All resulted in:
image

z0al: Ah, I see. If I get that right, I need to change the discard module to run something like the following in case of renames, correct?

```
git mv new_file old_file
```

altsem: I think so!

z0al: Fixed. The discard command now handles renames based on my testing.

z0al: Testing this again. If there are other changes in addition to renaming a file those changes won't get reverted. 

I think we need to run the checkout command in addition to `rename_file`

z0al: I never used the original `magit` so I'm not how it behaves in that case. Does it require two discards (first for the rename and another for the content changes)?

altsem: Interesting, I'd test this out in Magit and just copy the behaviour if possible.

I realised there may be other OpTrait implementations that may bork with this new "renamed" status.

The trait is found here:
https://github.com/altsem/gitu/blob/master/src/ops/mod.rs#L30

It's in the "TargetData" parameter.
When the TargetData is of the "Delta" variant.

I'd help test things out in Magit but I'm off the computer for a few days.

z0al: I just installed Emacs for the first time lol. Magit does one revert at a time. First, it reverts the renaming, and then you have to press `k` again to revert the other changes in the file. That matches the current behavior in `gitu` so I think we are good.

src/ops/discard.rs
```diff
@@ -49,6 +50,18 @@ fn clean_file(file: PathBuf) -> Action {
     })
 }
 
+fn rename_file(src: PathBuf, dest: PathBuf) -> Action {
+    Rc::new(move |state, term| {
+        let mut cmd = Command::new("git");
+        cmd.args(["mv", "--force"]);
```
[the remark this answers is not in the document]
altsem: I suppose Magit might prompt before moving to an already existing file. I think it's fine to keep --force, just like discard!

src/git/mod.rs
altsem: Interesting! I've never heard of it myself. Let's leave it :)

src/tests/snapshots/gitu__tests__discard__discard_file_move.snap
altsem, line 1: This snapshot is referring to this test:
https://github.com/altsem/gitu/blob/a30c2aea8873d1fbf025e80c7355049b56db0f37/src/tests/discard.rs#L39-L47

My guess is that it needs updating to do less keystrokes in order to discard a moved file now!

`y` is bound to "Copy commit hash", which is why it's seen at the end.

You could remove all key inputs in the test:
```
snapshot!(ctx, "");
```

And then see the diff by running:
```
cargo test discard_file_move
```

Then fill in as you see fit until it does what is expected!

altsem: (this TODO looks like it's old now too!)

altsem: (bonus points for adding more tests to stage/unstage moved files) :)

altsem (changes requested):
Looks pretty good I think.

But:
- There's a test-snapshot/test that needs fixing.

- I updated this README to better explain how commit messages work alongside the changelog generation.

   These commits
   - [style: align status columns](https://github.com/altsem/gitu/pull/234/commits/a16d6f8d89c3687df265b87519ae9470aa248675)
   - [fix: properly discard of renamed files](https://github.com/altsem/gitu/pull/234/commits/db2151212ff6de402894944ebb0ce259b0c0056c)

   .. would end up in the changelog, which may not be something that we want.

altsem (approved):

altsem:
@z0al i'll merge and release this. Gj! 🎉

PR #247, opened 2024-09-02
change priority order of editor envvar lookup

### LINKED ISSUE #246: Allow different editor for commit messages and opening files
I have my `$GIT_EDITOR` set to `code -rw`, which waits until the file is closed in VS Code. This is good for editing commit messages, but not for opening files.

It would be nice if we can configure a different editor for opening files in staged changes, untracked files, etc. Or maybe it'd be more sensible to use `$EDITOR` here instead?

eekhof:
I changed the order of priorities for the lookup of environment variables determining the editor to use for opening files or hunks from the main menu.

It was `GIT_EDITOR`, `VISUAL`, `EDITOR` and I changed it to `VISUAL`, `EDITOR`, `GIT_EDITOR`.

The rationale for this is that one might want to use a different editor for editing e. g. commit messages (which is set by `GIT_EDITOR`) than for opening/editing the changed code, which is what these environment variables are checked for. To allow this any one of `VISUAL` or `EDITOR` should take precedence.

This resolves #246 .

This is my first ever git contribution to another project, I am very sorry if I made any grave mistakes or oversights.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ A help-menu can be shown by pressing the `h` key, or by configuring `general.alw
 <img style="width: 720px" src="vhs/help.png"/>
 
 ### Configuration
-The environment variables `GIT_EDITOR`, `VISUAL` or `EDITOR` (checked in this order) dictate which editor Gitu will open.
+The environment variables `VISUAL`, `EDITOR` or `GIT_EDITOR` (checked in this order) dictate which editor Gitu will open. This means that e. g. commit messages will be opened in the `GIT_EDITOR` by Git, but if the user wishes to do edits to the actual files in a different editor, `VISUAL` or `EDITOR` can be set accordingly.
 
 Configuration is also loaded from:
 - Linux:   `~/.config/gitu/config.toml`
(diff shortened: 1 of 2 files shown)


altsem:
Initially thought of just dropping GIT_EDITOR, but changing the order seems better!

Just two things, 

The change log is automatically generated from commit messages, and it's be good if there was a descriptive message for when I make a release. See https://github.com/altsem/gitu/blob/master/docs/dev-tooling.md

I remember these variables are mentioned in the readme. Perhaps it'd be nice with an example and explanation of using them like this?

eekhof:
Perfect, thanks!

I tried to fix my first commit message and update the readme accordingly - How detailed would you like me to be in the example? I could give a concrete use case within Neovim as mentioned in #246 , but I cannot gauge if this is too specific for the general readme or not.

altsem (approved):
Looks great imo. I'll make a release and start using this myself. Thanks! :partying_face:

eekhof:
Wonderful, thank you! :)

Issue #58, opened 2024-03-12
[Feature request] Support upstream/pushRemote workflow


rynoV:
Thanks for making this! I really like it, especially for when I need to work on Windows.

One of the features I like about Magit is its support for easily differentiating a branch's [upstream/merge](https://git-scm.com/docs/git-config#Documentation/git-config.txt-branchltnamegtmerge) and [pushRemote](https://git-scm.com/docs/git-config#Documentation/git-config.txt-branchltnamegtpushRemote). I first saw the feature+workflow described [here](https://magit.vc/manual/magit/The-Two-Remotes.html), and now I find other git UIs without this feature like vscode's to be lacking. It's very convenient to be able to quickly see *both* the difference between the upstream merge target (e.g. `master`) and the remote branch being tracked (usually a "feature" branch with the same name as the local branch). For example, with Magit when I run a fetch:

- I can immediately see if some other feature has been merged to `master` and my feature branch has fallen behind
- I can immediately see if someone else, working on the same feature as me, has pushed to my feature branch and it has fallen behind
- I can quickly pull either `master` or the remote tracked branch into my local branch

The main features from Magit that I'd like to see in this project:

- Support for the separate push/fetch/pull keybinds described [here](https://magit.vc/manual/magit/The-Two-Remotes.html) for pushRemote vs upstream/merge
- Support for the two sections in the main status screen showing the diff between the local branch and pushRemote, and the local branch and upstream/merge
- Support for automatically setting `branch.<name>.pushRemote` (if it's not already set) when using the "push to pushRemote" keybind, in the same way as Magit

I'm not sure how big of an ask this is, I haven't looked at the codebase yet, but I like the project and I like rust so I might try implementing this at some point.

altsem:
Sounds great! In case you want to start on this, I'm happy to help!

For the status page, you'll find everything in [src/screen/status.rs](https://github.com/altsem/gitu/blob/master/src/screen/status.rs)

It produces a `Vec<Item>`, each Item being a visual line seen in Gitu, as well as some metadata.

rynoV:
I started looking into this today, I'm currently trying to figure out a good way to start experimenting with git2, planning to start setting up some tests using the `tests/helpers.rs` file. Glad to see there's already some good testing infrastructure set up!

rynoV:
[...]

PR #267, opened 2024-10-05
Configure/use upstream/pushRemote when pushing and pulling

altsem:
@rynoV I've been hacking around and testing things out. I opened up this new PR, but it continues on your work.
I'd appreciate any input on the code! :)

To summarize:
- Change bindings to `u` and `p`:
  `p` was previously the `push.default` (which often would be `upstream`).
  Now `u` is `upstream`, and `p` is `pushRemote`
- Fix when upstream was a local branch
- Make push-commands more explicit with refspecs (I copied Magit)
- Display `upstream` & `pushRemote` dynamically in the menus:
  `to origin/main` instead of `to upstream`

relates to: #58

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -334,12 +334,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "convert_case"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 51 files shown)


rynoV:
LGTM! Thanks for ironing out the rough edges like the refspecs and edge cases. I only briefly scanned your commits as I don't have much bandwidth for this atm, but I appreciate you looping me in and making the changes very easy to grok

Issue #124, opened 2024-04-11
Alias the `h` key to `?`


frank-lenormand:
Hi,

I intuitively hit <kbd>?</kbd> whenever I want to check a quick help reference.

Maybe that’s because I use <kbd>h</kbd>, <kbd>j</kbd>, <kbd>k</kbd>, <kbd>l</kbd> for actual operations in other terminals tools (like my editor).

Would it be possible to alias the <kbd>h</kbd> key to <kbd>?</kbd>?

altsem:
The default binding is `general.root.help_menu = ["h"]`. You could map it to both by doing:
`general.root.help_menu = ["h", "?"]`.

I would, but considering if `?` should be mapped to reverse search if `/` is mapped to search forward.

frank-lenormand:
I didn’t realise I could set this in the user configuration, thanks!

Issue #5, opened 2024-02-20
Search


altsem:
Search down/up in any screen.
Bind to `/` and `?`, `n` and `N` to go to next/prev?

`n` is currently bound to moving the cursor down. May want to bind `?` to help.
Does it make more sense to have separate vim/emacs-like keybind profiles?

mattgallagher92:
This would be great!

My guess is that people who prefer Emacs-like keybindings are likely to use Emacs and prefer to stay in Emacs, hence use Magit rather than gitu. So I'd argue that it makes sense to focus on making a good workflow for those using vim-style keybindings (which seems to be gitu's default). In particular, I'd support removing the `p` and `n` bindings to select previous/next to enable them to be used for other things.

(Note: I use vim-style keybindings for code editing and I'm a Magit user who basically only uses Emacs - with Evil mode for vi-style keybindings - for Magit, so I bring some hefty bias :sweat_smile:)

That said, it's going to be hard to get anything perfect - I'd love to be able to use `h` and `l` for fine-grained text selection when in visual mode (for example to search for a commit by message, then copy something like the hash), but there are obvious conflicts with help and log actions. You could get around that with full-on modes like Vim has, but I imagine that adds significant implementation complexity, and increases the barrier to entry for those not used to Vim. Certainly not worth consideration in the short term.

altsem:
I think dropping n/p for navigation might be reasonable. In the future, the keybinds should be configurable anyway.
Been thinking about adding a "copy hash" keybind somewhere.

abitrolly:
How popular is `?` for search? I use `/` and `Ctrl+F`, but never `?`. Most search just wraps around, so I very rarely need to do backward search with the shortcut. Using `?` for help, however, ever implemented here in GitHub.

Profpatsch:
Search is the one feature I need the most tbh, It’s really hard to work effectively without it.

PR #280, opened 2024-11-11
Make ? open help too, and close it also with ? or h

abitrolly:
Add ? for invoking help, and let the same key close it.

Fixes https://github.com/altsem/gitu/issues/124

### DIFF
--- a/src/default_config.toml
+++ b/src/default_config.toml
@@ -94,8 +94,8 @@ root.stage = ["s"]
 root.unstage = ["u"]
 root.copy_hash = ["y"]
 
-root.help_menu = ["h"]
-help_menu.quit = ["q", "<esc>"]
+root.help_menu = ["h", "?"]
+help_menu.quit = ["q", "h", "?", "<esc>"]
 
 root.branch_menu = ["b"]
 branch_menu.checkout = ["b"]


altsem:
Hey @abitrolly!

If think this'd make sense, if it were not for the plans to later add search functionality - which is commonly bound to `/` and `?`.

As found in: https://github.com/altsem/gitu/issues/5

abitrolly:
@altsem left some `?` binding related comments in https://github.com/altsem/gitu/issues/5#issuecomment-2471565549

altsem:
I did some googling. I think Doom Emacs uses `?` for the help menu (haven't confirmed this).
Is that the case for other evil-mode ways of using Magit?

If that is the case, it feels like a no-brainer to copy this behaviour.

abitrolly:
@altsem Magit itself uses `?` - https://github.com/niyumard/Doom-Emacs-Cheat-Sheet?tab=readme-ov-file#magit

altsem:
Let's merge this then :)
I noticed h -> h or ? -> ? opens up a manual in Doom. But we don't have one yet, so seems fine.

Thanks for talking some sense into me.

PR #291, opened 2024-12-15
Fix staircased git output

### LINKED ISSUE #290: Wrong output indentation
`gitu` show git output with wrong indentation

20241215_12h52m56s_grim

which is quite different from the `git` one:

20241215_12h53m07s_grim

This behavior can be reproduced in both bash and sh.

Platform: Nixos unstable and Termux
Version: both git main and v0.26

xubaiwang:
Fix #290 

This PR disables raw mode temporarily in interactive git command, which will fix the staircase git output caused by LF-CRLF conversion in raw mode.


altsem:
Nice catch! I've been bothered by this myself. :) 

Just two things I think:
- The commit title goes straight to the changelog. So it should be on a bit higher level. Something like "fix: staircased git output" perhaps?

- You could probably put enable/disable raw mode in the "Term" struct. `term.enable_raw_mode()?;` to fix the tests.

xubaiwang:
Just updated the commit. 

I have succeeded in running `cargo test` and `cargo insta test`. But `make test` still complains about other parts of the code.

altsem (approved):
Seems to work in CI! I'll merge this. Gj!

edit: oh the Linux CI fails whereas windows/mac succeed.
Looks like the linter had some rules added. It's a master-problem anyway.

PR #292, opened 2024-12-17
feat: update on file changes

sbillig:
This adds a file watcher (using the `notify` crate); when any files have been modified/created/removed/etc in the work directory, we do a screen update and redraw. I found gitu today and have been using it in zed as a magit substitute, and am really liking it so far (thanks!). I thought it would be nice to not have to manually refresh gitu.

The gitignore filter probably isn't necessary; if an ignored file is modified and triggers an extra update/redraw, it doesn't really matter. Or it could be more complete (eg I don't think we're currently filtering out files in `.git/` for example). 🤷

Happy to make changes or whatever as you see fit.

https://github.com/user-attachments/assets/f6f1df3c-08f3-4ee5-a53c-ce5d91513690


altsem:
Thats cool! I read through it and think it makes sense.

I haven't tested it myself, or read the code in-depth. (I'm on vacation until February).

Could be a configurable option.

I wonder if it works on windows/mac?

I'd wish for the cursor to remain at the selected line post updates ):, though, that is a separate problem.

sbillig:
(Rebased on #293)

micahscopes:
Tested this on Linux and it works great!  Just need to fix a little glitch that happens when `event.paths` is empty.

> I'd wish for the cursor to remain at the selected line post updates

Looks like the cursor remains at the literal selected line number, which may have different information on it after the update.  So I guess at each notify event if you were to calculate how many lines have been added/removed you could jump the cursor by that number of lines in the opposite direction maybe?

sbillig:
> Tested this on Linux and it works great! Just need to fix a little glitch that happens when `event.paths` is empty.

Whoops, that paths[0] was a silly testing thing I forgot to change. Fixed, thanks!

altsem:
@sbillig what do you think of making it possible to disable this via config? Enabled per default.
Just in case someone has a preference, or is experiencing issues.

There's a module "config.rs" I believe.

altsem:
@sbillig what do you think of making it possible to disable this via config? Enabled per default.
Just in case someone has a preference, or is experiencing issues.

There's a module "config.rs" I believe.

sbillig:
> @sbillig what do you think of making it possible to disable this via config? Enabled per default. Just in case someone has a preference, or is experiencing issues.
> 
> There's a module "config.rs" I believe.

Yeah, I'm up for that. I'd like to improve the file filtering as well. I'll get back to this PR sometime this week.

sbillig:
@altsem I added a config option: `refresh_on_file_change.enabled = true`

altsem (approved):
Code looks good imo.
I'll merge this to master for now. And will look at making a release when I'm back in feb.

Gj @sbillig!

PR #297, opened 2024-12-26
feat: ctrl-c to quit the app

### LINKED ISSUE #258: `<ctrl+c>` to exit the application
Gitu should probably exit right away when pressing Ctrl+C

bloznelis:
Resolves #258.

Firstly, thanks for this - I love magit and command line, so this feels like a perfect tool for me. As a thank you, I figured I will contribute a small ticket for you.

Now `ctrl-c` will exit the app bypassing any default or custom bindings. If you want to tweak the logic (for example exit only when the amount of screens is 1 i.e. root-level) - add more conditions to `is_system_quit` function.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -12,6 +12,7 @@ use arboard::Clipboard;
 use crossterm::event;
 use crossterm::event::Event;
 use crossterm::event::KeyCode;
+use crossterm::event::KeyEvent;
 use crossterm::event::KeyEventKind;
 use crossterm::event::KeyModifiers;
 use git2::Repository;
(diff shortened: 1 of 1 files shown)


altsem:
Hi! This is great :)

I wonder if whether someone would ever want this to be configurable. 🤔

bloznelis:
>I wonder if someone would ever want this to be configurable. 🤔

Unlikely (?).

Now that I gave this a second though, I think that the initial implementation I mande is a bit naive. It wouldn't work if for some reason previous `update` step hangs/takes too long and the user wants to kill the app, because it would need to wait for previous step to finish before handling the ctrl-c.

A better way would be to react on signals. So in the case of `SIGINT` (produced by ctrl-c), we would interrupt and close the app without waiting for `update` to resolve.

bloznelis:
Probably should happen somewhere around [here](https://github.com/altsem/gitu/blob/master/src/lib.rs#L102:L112). If we receive the `SIGTERM` we break the `while` loop.

altsem:
Yea that'd make sense :)

If there's a pending external command running, perhaps the signal should be propagated to it.

I can't recall if Ctrl-C produces a signal when terminal raw-mode is enabled, might be a gotcha.

Issue #215, opened 2024-07-05
Getting "Standard input is not a terminal" with nano


m0nhawk:
I am not sure how to configure this.

When `EDITOR` is configured to nano, gitu is failing to open tracked/untracked files:

```
$ nano changed_file.txt
! Couldn't open editor nano due to: exited with code: 1
```

And after exiting gitu:

```
Standard input is not a terminal
```

---

I was able to narrow it down to:

```
$ echo "echo" | nano
Standard input is not a terminal
$ echo "echo" | nano -
# nano editor
```

This is happening with `GNU nano, version 8.0`, but, for example, works with default "nano" (which is pico) in macOS: `Pico 5.09`.

Happening in both fish and zsh, so to me it looks like an issue with how gitu sends input to nano.

the-kenny:
I'm still seeing this issue on macOS on [0f02879](https://github.com/altsem/gitu/commit/0f02879be30ccd6dfe2bcf6a66a0e7958ee9d20c) (build locally). Some additional behaviour details:

`nano`, `zile`  show no cursor, but receive normal keystrokes (characters, etc.) but no control events (`up/down, ctrl-x, ...`). `vim` shows the behaviour described by @m0nhawk,  `emacs` crashes immediately (unfortunately I'm not able to get the error output).

the-kenny:
This may be fixed by using `Stdio::inherit()` when spawning in `run_interactive`. Can anyone affected by this try a build from my branch to confirm the fix is working for them?

https://github.com/the-kenny/gitu/tree/fix-215-terminal-input-to-editor

m0nhawk:
I works for me:

```
gitu on  fix-215-terminal-input-to-editor via 🦀 v1.83.0 took 34s λ touch 111
gitu on  fix-215-terminal-input-to-editor [?] via 🦀 v1.83.0 λ cargo run
   Compiling gitu v0.27.0 (/home/resu/src/gitu)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.88s
     Running `target/debug/gitu`
```

And `master` branch doesn't with this error:
```
gitu on  fix-215-terminal-input-to-editor [?] via 🦀 v1.83.0 took 9s λ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
gitu on  master [?] via 🦀 v1.83.0 λ cargo run
   Compiling gitu v0.27.0 (/home/resu/src/gitu)
[...]

PR #301, opened 2025-01-05
Use `Stdio::inherit()` in `run_cmd_interactive` (likely fix for #215)

the-kenny:
Likely fix for #215. `Command` is configured to `Stdio::pipe()` input to the spawned subprocess, likely causing the issues.

This commit also unhides the cursor before spawning the command, fixing hidden cursor issues when `git` spawns `EDITOR`.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -305,6 +305,8 @@ impl State {
         // Previously `Stdio::pipe()` was used here, likely causing
         // https://github.com/altsem/gitu/issues/215 and similar issues.
         cmd.stdin(Stdio::inherit());
+        // Redirect stderr so we can capture it via `Child::wait_with_output()`
+        cmd.stderr(Stdio::piped());
 
         // git will have staircased output in raw mode (issue #290)
         // disable raw mode temporarily for the git command
--- a/src/state.rs
+++ b/src/state.rs
@@ -302,11 +302,17 @@ impl State {
 
         cmd.current_dir(self.repo.workdir().expect("No workdir"));
 
-        cmd.stdin(Stdio::piped());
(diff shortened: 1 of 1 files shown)


altsem (changes requested):
Much needed fixes. Thanks!

src/state.rs
```diff
@@ -302,12 +302,18 @@ impl State {
...
         // disable raw mode temporarily for the git command
         term.backend().disable_raw_mode()?;
 
+        // If we don't show the cursor prior spawning (thus restore the default
+        // state), the cursor may be missing in $EDITOR.
+        term.show_cursor()?;
+
         let child = cmd.spawn()?;
 
         let out = child.wait_with_output()?;
```
altsem, line 319: This line is supposed to get the output of the command, which is then displayed at the bottom pop-up in Gitu.

I wonder if its behaviour would change with "Stdio::inherit".

We'd like to capture and display error messages from commands like "git commit".

the-kenny: A simple test (`c c`, then leaving commit message empty) shows `Aborting commit due to empty commit message.`.

We only change the behaviour for `stdin`, leaving `stderr` and `stdout` as-is. I'll push an update that makes this explicit by setting `.stdout(..)` and `.stderr(..)`.

the-kenny: [9f2d9ab](https://github.com/altsem/gitu/pull/301/commits/9f2d9ab2fecf82ef1cf162223abb17e86d34fe4e) now explicitly captures `stderr` (but not `stdout`, as it is necessary for the CLI editors to be visible) to capture the command's output.

altsem:
Following up on this in https://github.com/altsem/gitu/pull/311

PR #324, opened 2025-03-01
Add support for `nvr` command with line number navigation

ttak0422:
### About

This PR adds support for the `nvr` command, allowing users to jump directly to a specific line number using the `+[num]` option.

https://github.com/mhinz/neovim-remote


altsem (approved):
Nice! :)

PR #329, opened 2025-03-02
fix: respect pushDefault config option

### LINKED ISSUE #322: Pushing does not respect pushDefault git option [correction]
Hello,

I came across the problem today that I now need to manually set every remote to push to when first opening a repo in gitu. I believe this was not the case earlier, it would just default to origin. (If there is an idea why this is the case I would also be thankful, but this is not my main point, it might as well be some setting I changed accidentally somewhere.)

Trying to fix this I set `git config --global remote.pushDefault origin`, which adds the setting
```
[remote]
	pushDefault = origin
```
to every `.git/config` in a repo. I assumed this would make gitu default to origin, so I would not have to specify when first pushing. This is because when I manually set the remote to origin, gitu adds
```
[branch "main"]
...
	pushRemote = origin

```
to the config. But I would expect gitu to respect the `pushDefault` if there is no `pushRemote` present, as outlined in the [git config documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-remotepushDefault).

However gitu still prompts me `Set pushRemote then push:` when trying to push, even if `pushDefault` is already set.

Is this intentional, or would it be beneficial to respect `pushDefault` when present, so as to not prompt for a `pushRemote`?

Thanks for your time.

eekhof:
This fixes #322 , so that if no remote to push to is set, the default option set by
```
[remote]
	pushDefault = origin
```
Is used instead.

### DIFF
--- a/src/git/remote.rs
+++ b/src/git/remote.rs
@@ -101,6 +101,17 @@ pub(crate) fn head_push_remote_cfg(repo: &Repository) -> Res<String> {
     } else {
         return Err("Head is not a branch".into());
     };
+
     let push_remote_cfg = format!("branch.{branch}.pushRemote");
-    Ok(push_remote_cfg)
+    let config = repo.config()?;
+    if config.get_string(&push_remote_cfg).is_ok() {
+        Ok(push_remote_cfg)
+    } else {
+        let push_default_cfg = "remote.pushDefault".to_string();
+        if config.get_string(&push_default_cfg).is_ok() {
+            Ok(push_default_cfg)
+        } else {
+            Err("Neither pushRemote nor push.default is configured".into())
+        }
+    }
 }


src/git/remote.rs
```diff
@@ -101,6 +101,17 @@ pub(crate) fn head_push_remote_cfg(repo: &Repository) -> Res<String> {
...
-    Ok(push_remote_cfg)
+    let config = repo.config()?;
+    if config.get_string(&push_remote_cfg).is_ok() {
+        Ok(push_remote_cfg)
+    } else {
+        let push_default_cfg = "remote.pushDefault".to_string();
+        if config.get_string(&push_default_cfg).is_ok() {
+            Ok(push_default_cfg)
+        } else {
+            Err("Neither pushRemote nor push.default is configured".into())
```
altsem, line 114:
```suggestion
            Err("Neither branch.<name>.pushRemote nor remote.pushDefault is configured".into())
```

altsem: push.default is something else (Defaults to "simple" etc.)

altsem:
I'm a bit confused, so had a look at the `man git config`.
```
       remote.pushDefault
           The remote to push to by default. Overrides branch.<name>.remote for all branches, and is
           overridden by branch.<name>.pushRemote for specific branches.

```
What does `Overrides branch.<name>.remote for all branches` actually mean? :thinking:

```
       branch.<name>.remote
           When on branch <name>, it tells git fetch and git push which remote to fetch from or push to.
           The remote to push to may be overridden with remote.pushDefault (for all branches). The
           remote to push to, for the current branch, may be further overridden by
           branch.<name>.pushRemote. If no remote is configured, or if you are not on any branch and
           there is more than one remote defined in the repository, it defaults to origin for fetching
           and remote.pushDefault for pushing. Additionally, . (a period) is the current local
           repository (a dot-repository), see branch.<name>.merge's final note below.

       branch.<name>.pushRemote
           When on branch <name>, it overrides branch.<name>.remote for pushing. It also overrides
           remote.pushDefault for pushing from branch <name>. When you pull from one place (e.g. your
           upstream) and push to another place (e.g. your own publishing repository), you would want to
           set remote.pushDefault to specify the remote to push to for all branches, and use this option
           to override it for a specific branch.

```

It feels like a can of worm has opened up since the introduction of the two remotes (upstream/pushRemote).

altsem (changes requested):
Some tests are failing (`cargo test`).
It's also possible to add a test like (in `src/tests/push.rs`):
```rust
#[test]
fn push_doesnt_prompt_for_remote_when_default_set() {
    let ctx = TestContext::setup_clone();
    run(ctx.dir.path(), &["git", "checkout", "-b", "new-branch"]);
    commit(ctx.dir.path(), "new-file", "");
    run(
        ctx.dir.path(),
        &["git", "config", "remote.pushDefault", "origin"],
    );

    snapshot!(ctx, "Pp<enter>");
}
```

altsem:
This was duplicated in https://github.com/altsem/gitu/pull/347.
It's merged now, so I'll close this.

PR #337, opened 2025-03-30
Fix nix build

ipsavitsky:
Nix build fails because nix environment contains cargo ver. 1.79.0 which is incompatible with some of the crates gitu is using. This change updates nix flake inputs, which brings in newer cargo compatible with all crates


altsem:
Cool, let's try it!

altsem (approved):

PR #341, opened 2025-04-12
show: Add `GITU_SHOW_EDITOR` env-var

the-kenny:
This allows users to set a custom editor used for 'show' (e.g. pressing RET on a file in the status buffer).

Please let me know if you would prefer a PR that adds this to the config file.

### DIFF
--- a/src/ops/show.rs
+++ b/src/ops/show.rs
@@ -50,7 +50,7 @@ fn goto_show_screen(r: String) -> Option<Action> {
     }))
 }
 
-pub(crate) const EDITOR_VARS: [&str; 3] = ["VISUAL", "EDITOR", "GIT_EDITOR"];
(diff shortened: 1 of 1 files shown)


altsem:
@the-kenny I think a corresponding config would make sense, but is not a requirement.

Perhaps they would be prioritised as:
1. `GITU_SHOW_EDITOR`
2. config value (if set)
3. ..the rest

the-kenny:
Sure - I think a config option would make sense. I’ll add it today or tomorrow.

altsem:
Seems the linter has been updated in CI. I'll merge this and fix that problem!

altsem (approved):

altsem:
@the-kenny I forgot, about commit messages. There's automatic changelog generation in CI. If you name your commits a certain way, it will show up in the changelog.

`git cliff --unreleased` to check what would be added.

I'll add a changelog entry to this already merged PR.

PR #344, opened 2025-04-16
Custom `show` editor: Add configuration option

the-kenny:
Follow up to #341

### DIFF
--- a/src/ops/show.rs
+++ b/src/ops/show.rs
@@ -50,7 +50,7 @@ fn goto_show_screen(r: String) -> Option<Action> {
     }))
 }
 
-pub(crate) const EDITOR_VARS: [&str; 3] = ["VISUAL", "EDITOR", "GIT_EDITOR"];
+pub(crate) const EDITOR_VARS: [&str; 4] = ["GITU_SHOW_EDITOR", "VISUAL", "EDITOR", "GIT_EDITOR"];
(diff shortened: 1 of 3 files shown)


src/ops/show.rs
```diff
@@ -50,16 +50,18 @@ fn goto_show_screen(r: String) -> Option<Action> {
     }))
 }
 
-pub(crate) const EDITOR_VARS: [&str; 3] = ["VISUAL", "EDITOR", "GIT_EDITOR"];
+pub(crate) const EDITOR_VARS: [&str; 4] = ["GITU_SHOW_EDITOR", "VISUAL", "EDITOR", "GIT_EDITOR"];
```
altsem, line 53: These env vars are mentioned here:
https://github.com/altsem/gitu/blob/ebb63fbb6a1a68618689070a7a99fd096ff135a8/README.md?plain=1#L33

GITU_SHOW_EDITOR could be included as well there.

altsem (changes requested):
Just two feedback points:
- I would expect GITU_SHOW_EDITOR to override editor.show.
- The README mention would be nice!

Sharing my thoughts, but not a requirement:
Another approach could even be to make any config value override-able with env vars. I haven't poked around with it yet.
Config is loaded here:
https://github.com/altsem/gitu/blob/ebb63fbb6a1a68618689070a7a99fd096ff135a8/src/config.rs#L175-L191

PR #347, opened 2025-04-16
feat: Fall back to remote.pushDefault when branch pushRemote is not set

### LINKED ISSUE #346: remote.pushDefault is not consulted before prompting for a pushRemote
This was previously reported in #321, although I can't quite figure out why it was closed.

My [remote.pushDefault](https://git-scm.com/docs/git-config#Documentation/git-config.txt-branchltnamegtremote) config option in my repo is set to `origin`:

```shell
git config remote.pushDefault
origin
```

But when I do `Pp`, I am still prompted to enter a `pushRemote` for every new branch.

Briefly looking at `get_push_remote` and `head_push_remote_cfg`, I couldn't see how it might fall back to `remote.pushDefault` if `branch.{branch}.pushRemote` was not defined.

(I'll admit I don't think I fully understand the multitude of ways git can find a push remote, so maybe there's some other config value I should have looked for first?)

jonathanj:
In the case where `get_push_remote` would have returned `Ok(None)` now first try `get_default_push_remote` which will consult the `remote.pushDefault` git config option, before either returning the fallback push remote, or no result.

Fixes #346

### DIFF
--- a/src/git/remote.rs
+++ b/src/git/remote.rs
@@ -77,6 +77,22 @@ pub(crate) fn get_push_remote(repo: &Repository) -> Res<Option<String>> {
                 .map_err(Utf8Error::String)
                 .map_err(Error::ReadGitConfigUtf8)?,
         )),
+        Err(e) if e.class() == git2::ErrorClass::Config => get_default_push_remote(repo),
+        Err(e) => Err(Error::ReadGitConfig(e)),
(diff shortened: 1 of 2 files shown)


altsem (approved):
Looks good! tests, and nice commit message. I'll just merge it.

PR #350, opened 2025-04-17
feat(branch): add delete branch functionality

rachartier:
Hello,

I've added code to delete a branch by its name, or one selected by branch from `Y`. 
Tell me what you think, and if I need to improve or change something.

Thanks for this amazing project.


src/ops/delete_branch.rs
```diff
@@ -0,0 +1,66 @@
...
+        return Err(Error::NotOnBranch);
+    }
+
+    // Don't allow deleting current branch
+    let current_branch = state
+        .repo
+        .head()
+        .map_err(Error::GetHead)?
+        .shorthand()
+        .unwrap_or("")
```
altsem, line 49: .shorthand will be None if it is invalid utf8 (It's great :disappointed:). And I think we should abort if so.

Here's an example of handling it more explicitly:
https://github.com/altsem/gitu/blob/947da1b8015f49f32a76c704b50871b4070a3ff7/src/git/mod.rs#L220-L227

Handling these is a bit messy, so the idea is the stack some helpers in src/git/mod.rs for now (like get_current_branch_name, get_head_name, get_current_branch).

src/ops/delete_branch.rs
```diff
@@ -0,0 +1,66 @@
...
+        .to_string();
+
+    if current_branch == branch {
+        state.display_error("Cannot delete the current branch".to_string());
+        return Ok(());
```
altsem, line 54: This is handled from the outside, better to define a new error and return in.

But `state.close_menu();` would then need to happen before.

Something like:
```suggestion
        return Err(Error::CantDeleteCurrentBranch);
```

src/ops/delete_branch.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    let delete_flag = "-d";
```
altsem, line 57: Git will block the deletion if the branch is unmerged, and Magit would prompt the user in this case.
I think including this as well would be good, a lot of users would run into it.

Here's what Magit would prompt:
```
Delete unmerged branch fix-pushdefault? (y or n)
```

src/ops/delete_branch.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    fn display(&self, _state: &State) -> String {
+        "Delete branch".into()
+    }
+}
+
+fn delete_branch(state: &mut State, term: &mut Term, branch: &str) -> Res<()> {
+    if branch.is_empty() {
+        return Err(Error::NotOnBranch);
```
altsem, line 40: Perhaps `NotOnBranch` isn't quite true in this case. Maybe define a new `InvalidBranch`?

src/menu.rs
```diff
@@ -50,7 +50,12 @@ impl PendingMenu {
             is_hidden: false,
             args: match menu {
                 Menu::Root => vec![],
-                Menu::Branch => ops::checkout::init_args(),
+                Menu::Branch => {
+                    // Merge arguments from checkout and delete_branch
+                    let mut args = ops::delete_branch::init_args();
+                    args.extend(ops::checkout::init_args());
+                    args
+                }
```
altsem, line 58: The "Branch menu" maps 1-1 to the src/ops/checkout.rs module.

And these args refer to the arguments seen in that menu.

Therefore I think the delete_branch module should be removed, and the action moved into the "checkout" module.

src/default_config.toml
```diff
@@ -101,6 +101,7 @@ help_menu.quit = ["q", "h", "?", "<esc>"]
 root.branch_menu = ["b"]
 branch_menu.checkout = ["b"]
 branch_menu.checkout_new_branch = ["c"]
+branch_menu.delete_branch = ["D"]
```
altsem, line 104: This is bound to 'x' in Doom's Magit. I think that should be the default.

altsem (changes requested):
Hi. Nice to see this addition!

I got some feedback. See if they make sense.
A few tests are needed too. I found the related ones here:
https://github.com/altsem/gitu/blob/947da1b8015f49f32a76c704b50871b4070a3ff7/src/tests/mod.rs#L260-L288

rachartier:
Does it seem better?

I've tried to implement all your suggestions. Thanks for the input!

Tell me what you think

src/default_config.toml
```diff
@@ -101,6 +101,7 @@ help_menu.quit = ["q", "h", "?", "<esc>"]
 root.branch_menu = ["b"]
 branch_menu.checkout = ["b"]
 branch_menu.checkout_new_branch = ["c"]
+branch_menu.delete_branch = ["x", "D"]
```
altsem, line 104: I did actually stumble upon running Magit with the default bindings.
And I noticed it is not bound to D.

```
Checkout                             Create           Do
 b branch/revision   c new branch     n new branch     C configure...
 l local branch      s new spin-off   S new spin-out   m rename
                                                       x reset
                                                       k delete
```

I'd probably vote against binding delete to `k` in Gitu, since it has vim-like bindings.
`K` is bound to `Discard` in other places, so might feel natural.

I'm a bit split whether we should
- Copy Doom Emacs and bind `x`.
- Try improve and make things more consistent, bind it to `K`.
- Bind both.

What do you think? I suppose you suggested `D` due to the CLI flag?

src/ops/checkout.rs
```diff
@@ -64,3 +67,60 @@ fn checkout_new_branch_prompt_update(state: &mut State, term: &mut Term) -> Res<
...
+                let branch = branch.clone();
+                Some(create_y_n_prompt(
+                    Rc::new(move |state, term| delete_branch(state, term, &branch)),
+                    "Delete this branch?",
+                ))
+            }
+            _ => Some(create_prompt_with_default(
+                "Delete branch",
+                delete_branch,
+                selected_rev,
```
altsem, line 85: selected_rev doesn't do the trick here. If my cursor is on a branch/commit in the log, then this'll prompt me to delete the commit hash.

src/tests/mod.rs
```diff
@@ -285,6 +285,28 @@ mod checkout {
...
+        snapshot!(ctx, "YjbDmain<enter>");
+    }
+
+    #[test]
+    pub(crate) fn delete_unmerged_branch() {
+        let ctx = TestContext::setup_clone();
+        run(ctx.dir.path(), &["git", "checkout", "-b", "unmergedbranch"]);
+        commit(ctx.dir.path(), "unmergedfile", "This is unmerged content");
+        run(ctx.dir.path(), &["git", "checkout", "main"]);
+        snapshot!(ctx, "YjjbDy<enter>");
```
altsem, line 308: Let's copy the flow of Magit.

1. Prompt for a branch (Default to selected item's branch)
2. User presses `<enter>` and branch is deleted, unless:
3. An additional prompt to confirm deletion of an unmerged branch is displayed if it's the case
4. User presses `y` and branch is deleted.

src/ops/checkout.rs
```diff
@@ -64,3 +67,60 @@ fn checkout_new_branch_prompt_update(state: &mut State, term: &mut Term) -> Res<
...
+        let target_branch = crate::git::get_branch(&state.repo, branch)?;
+        if crate::git::is_branch_unmerged(&state.repo, &target_branch).unwrap_or(false) {
+            "-D"
+        } else {
+            "-d"
+        }
+    };
+
+    let mut cmd = Command::new("git");
+    cmd.args(["branch", delete_flag, branch]);
```
altsem, line 122:
```suggestion
    cmd.args(["branch", "-D", branch]);
```

I think this'll work too!

src/ops/checkout.rs
```diff
@@ -64,3 +67,60 @@ fn checkout_new_branch_prompt_update(state: &mut State, term: &mut Term) -> Res<
...
+impl OpTrait for DeleteBranch {
+    fn get_action(&self, target: Option<&TargetData>) -> Option<Action> {
+        match target {
+            Some(TargetData::Branch(branch)) => {
+                let branch = branch.clone();
+                Some(create_y_n_prompt(
+                    Rc::new(move |state, term| delete_branch(state, term, &branch)),
+                    "Delete this branch?",
+                ))
+            }
```
altsem, line 81: Let's remove this case, and just disregard `target`. I explained it further down below in the test. :)

src/tests/mod.rs
altsem, line 308:
```suggestion
        snapshot!(ctx, "YjjbDy");
```

As of how it works right now, the final <enter> is redundant. Let's change the behaviour though!

src/tests/mod.rs
```diff
@@ -285,6 +285,28 @@ mod checkout {
     pub(crate) fn checkout_new_branch() {
         snapshot!(TestContext::setup_clone(), "bcf<esc>bcx<enter>");
     }
+
+    #[test]
+    pub(crate) fn delete_branch_menu() {
+        let ctx = TestContext::setup_clone();
+        run(ctx.dir.path(), &["git", "branch", "branch-to-delete"]);
+        snapshot!(ctx, "YjbDy");
+    }
```
altsem, line 294: I think a lot of other similar tests are structured like:

```suggestion

    #[test]
    pub(crate) fn delete_branch_prompt() {
        let ctx = TestContext::setup_clone();
        run(ctx.dir.path(), &["git", "branch", "branch-to-delete"]);
        snapshot!(ctx, "YjbD");
    }

    #[test]
    pub(crate) fn delete_branch_menu() {
        let ctx = TestContext::setup_clone();
        run(ctx.dir.path(), &["git", "branch", "branch-to-delete"]);
        snapshot!(ctx, "YjbDy");
    }
```

src/ops/mod.rs
```diff
@@ -10,6 +10,7 @@ use std::{fmt::Display, rc::Rc};
 pub(crate) mod checkout;
 pub(crate) mod commit;
 pub(crate) mod copy_hash;
+
```
altsem, line 13:
```suggestion
```

altsem (changes requested):
Hey. Looks pretty good. I did some testing and found some things still. Good job though. Let's make it great :)

src/default_config.toml
rachartier: I'm a Neogit user, and it uses `D` as the command to delete.
image

That's why I used it.

I think it is better to use a unified method throughout the entire project, as you said it might feel more natural. So `K` is probably the right choice here.

src/tests/mod.rs
rachartier: I cannot, for the life of me, use `create_y_n_prompt`. I'm still new to Rust, but I can't seem to deal with these errors:
```
Diagnostics:
1. cannot borrow data in an `Rc` as mutable
   trait `DerefMut` is required to modify through a dereference, but it is not implemented for `std::rc::Rc<dyn for<'a, 'b> std::ops::FnMut(&'a mut state::State, &'b mut ratatui::Terminal<term::TermBackend>) -> std::result::Result<(), error::Error>>` [E0596]
```
I've tried for about 2 hours, but I cannot resolve this issue.
Could you please provide some guidance on what to do?

You can look at the code here: https://github.com/rachartier/gitu/blob/f99c26117778b835f765a58777cfc9aec6836c1a/src/ops/checkout.rs#L119

altsem: Took me some time to figure this out before too.
There's a function to get a mutable reference to the value within an Rc, it only works when there's a single reference (as in this case).

You have to call these functions like: `Rc::get_mut(..)`.
https://doc.rust-lang.org/std/rc/struct.Rc.html

```rust
        let mut prompt = create_y_n_prompt(action, "Branch is unmerged. Really delete?");
        Rc::get_mut(&mut prompt).unwrap()(state, term)?;
```

rachartier: Thanks! I did try that, but then the menu will not update. Did I forget something?


https://github.com/user-attachments/assets/452591ae-dd65-42a6-a27c-6f0e8c18a2a9

altsem: Trying to set my mind into this again, it's quite messy.
I didn't solve it, but here's what I found:
- create_y_n_prompt returns a function just that runs `state.prompt.set(..)`.
  https://github.com/altsem/gitu/blob/ebb63fbb6a1a68618689070a7a99fd096ff135a8/src/ops/mod.rs#L211-L218
- The prompt is later updated in the main loop:
  https://github.com/altsem/gitu/blob/ebb63fbb6a1a68618689070a7a99fd096ff135a8/src/state.rs#L129
  https://github.com/altsem/gitu/blob/ebb63fbb6a1a68618689070a7a99fd096ff135a8/src/state.rs#L147-L167
- Perhaps there should be a way to chain callbacks, handing over control to the main loop in-between.

altsem: @rachartier I refactored the prompt functions a bit to try simplify (it's in master).
After having thought of this more, I think I am more convinced that:
- There should be a function to relinquish control to the user (and get a result perhaps?).
  ```rust
  let response = prompt(state, term, "How are you");
  ```

I'll keep experimenting!

rachartier: Sound really nice !

altsem: Work in progress at https://github.com/altsem/gitu/pull/356

rachartier: Do you want me to pull your modifications and try deleting the branch? Or should I wait until you merge into main?

altsem: I think it's easiest if you wait, I might amend the commit.
Or feel free to try it if you want to. (If you merge it, it's easier to undo perhaps).

altsem: @rachartier I made some more progress on that separate branch. It just needs some cleanup. Got the implementation/tests working! :partying_face:. I'll find a day soon again to do the rest.

altsem: @rachartier changes are in master! If you rebase, you'll find things a bit easier to work with. No need for callbacks!

```rust
state.confirm(term, "Are you sure? (y or n)")?;

let input = state.prompt(
    term,
    &PromptParams {
        prompt: "Enter some input",
        create_default_value: Box::new(selected_stash),
        ..Default::default()
    },
)?;

do_something(input);
```

rachartier:
Thank you a lot for your prompt modifications. It made it a whole lot easier! 

I've made a new version that you can review.

src/ops/checkout.rs
```diff
@@ -80,3 +81,76 @@ fn checkout_new_branch_prompt_update(
...
+
+        let target_branch = crate::git::get_branch(&state.repo, branch)?;
+
+        crate::git::is_branch_unmerged(&state.repo, &target_branch).unwrap_or(false)
+    };
+
+    if is_unmerged {
+        let branch_to_delete = branch.to_string();
+
+        match state.confirm(term, "Branch is unmerged. Really delete?") {
```
altsem, line 133:
```suggestion
        match state.confirm(term, "Branch is unmerged. Really delete? (y or n)") {
```

Ugly, but...
Let's throw in a `(y or n)` to make it consistent with the other ones? I should've just taken the effort to bake it into `state.confirm`, perhaps later.

altsem (changes requested):
It looks pretty good! I have only two requests:

### Consistency with "Discard" action
I found some more code while testing it out. This already existing action doesn't support deleting unmerged branches like the new one. It feels fine to have both of them, two different entry-points. `K` is always at-the-cursor, `bK` would prompt for a branch.
But perhaps they should share the underlying implementation?
https://github.com/altsem/gitu/blob/master/src/ops/discard.rs#L46-L57

Found here:
https://github.com/altsem/gitu/blob/master/src/ops/discard.rs#L46-L57

### The generated changelog is off
```
gitu on  rachartier/master:master ❯ git cliff --unreleased
# Changelog

All notable changes to this project will be documented in this file.

## [unreleased]

### 🚀 Features

- Expose selected branch in state
- *(checkout)* Prompt before deleting unmerged branch
- *(config)* Add 'x' as default keybindings for branch operations
- *(branch)* Add delete branch functionality
```
see https://github.com/altsem/gitu/blob/master/docs/dev-tooling.md

Some ideas:
```
feat(branch): Deleting (unmerged) branches via the branch menu
feat(discard): Support for deleteing unmerged branches
```

altsem:
This was picked up in https://github.com/altsem/gitu/pull/367, I'll close this.
Gj though, we laid out the ground for it.

PR #352, opened 2025-04-18
feat: Switch terminal backend from Crossterm to Termwiz

altsem:
The goal is to inevitably switch from Ratatui, and start using Termwiz
with Termwiz widgets.

My hope is that this will clean up the UI code quite a bit,
as well as allow more features like re-interpreting terminal codes,
and perhaps using Termwiz's line editor with autocomplete for prompts.

This merely switches the "backend" of Ratatui to Termwiz.

TODO:
- [ ] Test on Mac
- [ ] Test on Windows
- [ ] Fix: hunks turn black when selected and extending beyond terminal
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
+version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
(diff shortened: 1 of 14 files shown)


altsem:
I can replicate this resizing bug too. And do recall that it was mentioned in an example I read:
https://github.com/wezterm/wezterm/blob/909573fad98a1cccb802a7e7e1d9f4f44e02a177/termwiz/examples/widgets_basic.rs#L119-L125

I didn't manage to figure out a fix for the rendering bug.
It seems to occur when there's a style set on the bottom-most right cell of the terminal.
Bug in Ratatui?
`~/.config/gitu/config.toml`
```
[style]
selection_area.bg = "10"
```

One alternative is to rewrite the entire rendering code in Termwiz straight away.

PR #360, opened 2025-04-28
feat: FileWatcher now ignores changes from patterns in .gitignore

### LINKED ISSUE #358: `FileWatcher` doesn't seem to work on macOS any more
Doing `git bisect` I ended up on https://github.com/altsem/gitu/commit/0fff20d2accb475c0000922ba00f5546e4defa25 as being the cause of it no longer working.

In my case, the `RecommendedWatcher::kind()` is `Fsevent`, and for reasons I don't understand, doing `watcher.watch` inside of the thread, doesn't seem to actually watch anything. I put a breakpoint inside of the `recommended_watcher` event handler to confirm this. I tried upgrading [notify](https://docs.rs/notify) to 8.0.0, and poked around the Github project but couldn't find any clues (it's possible I did a bad job of looking.) The documentation for `notify` doesn't mention using `thread::spawn`, maybe this is obvious to a seasoned Rust programmer, but I was expecting to find some advice in the docs.

I'm not sure what the issue was that lead to the previously mentioned commit, perhaps it was `PollWatcher` causing a hang? Maybe on another OS? Maybe in a specific repo setup?

What I haven't tried yet is manually creating `PollWatcher` or `KQueue` on macOS to see if they still work when used from a thread.

### LINKED ISSUE #359: `FileWatcher` doesn't ignore gitignored files
(NOTE: I ran across this while investigating https://github.com/altsem/gitu/issues/358, without `FileWatcher` working the `event_handler` function obviously won't be invoked.)

Running with `--log` and then doing `cargo build` produces this output, which I would have expected to be ignored given [the contents of `.gitignore`](https://github.com/altsem/gitu/blob/740a86f1909744e59380aa990c1409e9324746bd/.gitignore).

```
--8<-- A couple dozen things changing in `target/` --8<--
[00:00:31.373] (16d8db000) INFO   File changed: "/Users/jonathan/Coding/gitu/target/debug/.fingerprint/gitu-80327dc8460d6252/test-bin-gitu"
[00:00:31.373] (16d8db000) INFO   File changed: "/Users/jonathan/Coding/gitu/target/debug/.fingerprint/gitu-80327dc8460d6252/test-bin-gitu.json"
[00:00:31.374] (16d8db000) INFO   File changed: "/Users/jonathan/Coding/gitu/target/debug/.fingerprint/gitu-80327dc8460d6252/test-bin-gitu.json"
[00:00:31.374] (16d8db000) INFO   File changed: "/Users/jonathan/Coding/gitu/target/debug/.fingerprint/gitu-80327dc8460d6252/test-bin-gitu.json"
[00:00:31.384] (16d8db000) INFO   File changed: "/Users/jonathan/Coding/gitu/target/.rustc_info.json"
[00:00:31.384] (16d8db000) INFO   File changed: "/Users/jonathan/Coding/gitu/target/.rustc_info.json"
[00:00:31.455] (200010240) DEBUG  FileUpdate
```

Investigating this, it seems like there are two contributing factors:

1. The `GitignoreBuilder` is not configured to call `.add(".gitignore")`, which `Gitignore::new` [does do](https://github.com/BurntSushi/ripgrep/blob/6dfaec03e830892e787686917509c17860456db1/crates/ignore/src/gitignore.rs#L110) (`path` there is the path to the gitignore, not to the root path as with `GitignoreBuilder::new`)
2. [`gitignore.matched`](https://github.com/altsem/gitu/blob/740a86f1909744e59380aa990c1409e9324746bd/src/file_watcher.rs#L35) only seems to naively match things, whereas [matched_path_or_any_parents](https://docs.rs/ignore/latest/ignore/gitignore/struct.Gitignore.html#method.matched_path_or_any_parents) will walk upwards through the parents, and eventually does match. I'm not sure I fully understand the ignore strategy automata stuff, and why `matched` is not intended to match `target/foo/bar.xyz` for a rule of `/target`, but `matched_path_or_any_parents` does work

```diff
diff --git a/src/file_watcher.rs b/src/file_watcher.rs
index 27dd9c3..b6e3efc 100644
--- a/src/file_watcher.rs
+++ b/src/file_watcher.rs
@@ -19,7 +19,10 @@ impl FileWatcher {
         let pending_updates = Arc::new(AtomicBool::new(false));
         let pending_updates_w = pending_updates.clone();
 
-        let gitignore = GitignoreBuilder::new(path)
+        let mut gitignore_builder = GitignoreBuilder::new(path);
+        // TODO: Handle `.add` error case.
+        gitignore_builder.add(".gitignore");
+        let gitignore = gitignore_builder
             .add_line(None, super::LOG_FILE_NAME)
             .map_err(Error::FileWatcherGitignore)?
             .build()
@@ -32,7 +35,10 @@ impl FileWatcher {
                 }
 
                 for path in event.paths {
-                    if !gitignore.matched(&path, path.is_dir()).is_ignore() {
+                    if !gitignore
+                        .matched_path_or_any_parents(&path, path.is_dir())
+                        .is_ignore()
+                    {
                         log::info!("File changed: {:?}", path);
                         pending_updates_w.store(true, Ordering::Relaxed);
                         break;
```

I'd be happy to make a PR, but I would need some guidance on how to correctly handle errors. The [documentation for add](https://docs.rs/ignore/latest/ignore/gitignore/struct.GitignoreBuilder.html#method.add) says that partial errors can be returned, which means that some rules can still have been added, so it seems non-fatal. An early exit here would mean the file watcher isn't actually started, which seems a bit drastic for some gitignore issues.

## Side note

I'm not sure what the issue was that caused https://github.com/altsem/gitu/commit/0fff20d2accb475c0000922ba00f5546e4defa25 to be necessary, but I wonder if the gitignore file not being honoured by the filewatcher could have contributed to the freezing issue.

jonathanj:
`FileWatcher` will build a gitignore ruleset from all `.gitignore` files in the repo and any global `.gitignore`, and use this to ignore file change events. This prevents a lot of `FileUpdate` events being generated when ignored items like build folders change, which may have been contributing to freezes.

If a change is detected involving a `.gitignore` file, the ruleset is rebuilt.

Changes are also now no longer watched from a thread, which caused problems with some watcher implementations such as `Fsevents` on macOS.

Fixes #359
Fixes #358

### DIFF
--- a/src/file_watcher.rs
+++ b/src/file_watcher.rs
@@ -1,29 +1,28 @@
-use crate::{error::Error, Res};
-use ignore::gitignore::GitignoreBuilder;
-use notify::{Event, EventKind, RecursiveMode, Watcher};
+use crate::{error::Error, open_repo, Res};
+use ignore::gitignore::{gitconfig_excludes_path, GitignoreBuilder, Gitignore};
+use notify::{Event, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
 use std::{
     path::Path,
+    path::PathBuf,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
     },
-    thread,
 };
 
 pub struct FileWatcher {
+    _watcher: RecommendedWatcher,
     pending_updates: Arc<AtomicBool>,
 }
 
 impl FileWatcher {
-    pub fn new(path: &Path) -> Res<Self> {
+    pub fn new(repo_dir: &Path) -> Res<Self> {
         let pending_updates = Arc::new(AtomicBool::new(false));
         let pending_updates_w = pending_updates.clone();
 
-        let gitignore = GitignoreBuilder::new(path)
-            .add_line(None, super::LOG_FILE_NAME)
-            .map_err(Error::FileWatcherGitignore)?
-            .build()
-            .unwrap();
+        let path_buf = repo_dir.to_owned();
+        let path_buf_copy = path_buf.clone();
+        let mut gitignore = build_gitignore(repo_dir)?;
 
         let mut watcher = notify::recommended_watcher(move |res: Result<Event, notify::Error>| {
             if let Ok(event) = res {
@@ -32,7 +31,20 @@ impl FileWatcher {
                 }
 
                 for path in event.paths {
-                    if !gitignore.matched(&path, path.is_dir()).is_ignore() {
+                    if path
+                        .file_name()
+                        .map_or(false, |name| name == ".gitignore")
+                    {
+                        log::info!("Rebuilding gitignore ruleset");
+                        if let Ok(new_gitignore) = build_gitignore(&path_buf_copy) {
+                            gitignore = new_gitignore;
+                        }
+                    }
+
+                    if !gitignore
+                        .matched_path_or_any_parents(&path, path.is_dir())
+                        .is_ignore()
+                    {
                         log::info!("File changed: {:?}", path);
                         pending_updates_w.store(true, Ordering::Relaxed);
                         break;
@@ -42,17 +54,18 @@ impl FileWatcher {
         })
         .map_err(Error::FileWatcher)?;
 
-        let path_buf = path.to_owned();
-        thread::spawn(move || {
-            if let Err(err) = watcher
-                .watch(path_buf.as_ref(), RecursiveMode::Recursive)
-                .map_err(Error::FileWatcher)
-            {
-                log::error!("Couldn't start file-watcher due to: {}", err);
-            }
-        });
+        watcher
+            .watch(&path_buf, RecursiveMode::Recursive)
+            .map_err(Error::FileWatcher)?;
+        log::info!(
+            "File watcher started (kind: {:?})",
+            RecommendedWatcher::kind()
+        );
 
-        Ok(Self { pending_updates })
+        Ok(Self {
+            _watcher: watcher,
+            pending_updates,
+        })
     }
 
     pub fn pending_updates(&self) -> bool {
@@ -65,4 +78,33 @@ fn is_changed(event: &Event) -> bool {
         event.kind,
         EventKind::Create(_) | EventKind::Modify(_) | EventKind::Remove(_)
     )
+}
+
+fn build_gitignore(path: &Path) -> Res<Gitignore> {
+    let mut gitignore_builder = GitignoreBuilder::new(path);
+    for gitignore_path in repo_gitignore_paths(path)? {
+        gitignore_builder.add(gitignore_path);
+    }
+    gitignore_builder.add_line(None, super::LOG_FILE_NAME).ok();
+    gitignore_builder
+        .build()
+        .map_err(Error::FileWatcherGitignore)
+}
+
+fn repo_gitignore_paths(repo_dir: &Path) -> Res<Vec<PathBuf>> {
+    let mut gitignore_paths = gitconfig_excludes_path().map_or_else(|| vec![], |path| vec![path]);
+    gitignore_paths
+        .extend(
+            open_repo(repo_dir)?
+                .index()
+                .map_err(Error::OpenRepo)?
+                .iter()
+                .filter_map(|entry| {
+                    match std::str::from_utf8(&entry.path).map(Path::new) {
+                        Ok(path) if path.file_name() == Some(std::ffi::OsStr::new(".gitignore")) => Some(path.to_path_buf()),
+                        _ => None
+                    }
+                })
+        );
+    Ok(gitignore_paths)
 }
--- a/src/file_watcher.rs
+++ b/src/file_watcher.rs
@@ -1,9 +1,7 @@
 use crate::{error::Error, open_repo, Res};
-use ignore::gitignore::{gitconfig_excludes_path, GitignoreBuilder, Gitignore};
 use notify::{Event, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
 use std::{
     path::Path,
-    path::PathBuf,
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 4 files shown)


jonathanj:
I've removed starting the watcher in a thread because I suspect that the cause was triggering a lot of update events as a result of should-have-been-ignored files changing (such as `target/`).

This is an optmistic change since I am not sure how to reproduce the freezes myself, so I can't say whether this fixes them or not. If there is a repo that will reproduce this, I'm happy to check it myself although I'll be using the `Fsevents` watcher on macOS.

https://github.com/altsem/gitu/issues/359#issuecomment-2832622563

jonathanj:
There are no tests because there are no existing tests for `FileWatcher` to extend, and I am really not sure how to write them from scratch. Happy to add them if you have a suggestion for how to approach it.

altsem:
Nice! I do agree we can drop spawning it in a thread and fixing things properly instead.
Tests would be cool I thought too, not sure where to begin. Perhaps if there was some way to refresh the UI only (without updating the state via `git diff`, `git status` etc.) in tests.

I believe this approach is missing some of the variants of ignoring stuff though, as seen in https://git-scm.com/docs/gitignore.

Did some digging and found this actually: https://git-scm.com/docs/git-check-ignore and https://docs.rs/git2/latest/git2/struct.Repository.html#method.status_should_ignore

And in gitu:
```
gitu on  359-filewatcher-apply-gitignore [!] ❯ git check-ignore target
target
gitu on  359-filewatcher-apply-gitignore [!] ❯ echo $status
0
```

Works in subdirs:
```
gitu/target on  359-filewatcher-apply-gitignore [!] ❯ git check-ignore .
.
gitu/target on  359-filewatcher-apply-gitignore [!] ❯ echo $status
0
```

I've disabled `showUntrackedFiles` in my git-managed home repo, it seems check-ignore returns false:
```fish
~ ❯ GIT_DIR="$HOME/repos/cfg.git" GIT_WORK_TREE="$HOME" git check-ignore .cache/gnome-calculator/un-daily.xls
~ ❯ echo $status # I'd hope for a 0 (ignored), but seems it doesn't consider showUntrackedFiles
1
~ ❯ GIT_DIR="$HOME/repos/cfg.git" GIT_WORK_TREE="$HOME" git config status.showUntrackedFiles
no
```

altsem:
But let's leave the tests aside. I believe it'd be messy, since file update events would come in async. And it'd need to have some sort of sleep duration in hope of catching it.

jonathanj:
> Did some digging and found this actually: https://git-scm.com/docs/git-check-ignore and https://docs.rs/git2/latest/git2/struct.Repository.html#method.status_should_ignore

Oops, in hindsight this seems like such an obvious thing to have looked for first, good find! If nothing else, this dramatically simplifies the ignore logic, to the point that I don't think the `ignore` crate is even necessary any more? I've pushed a change that switches to this and removes the `ignore` lib.

Aside, I noticed (after adding the event type for watch events) that some of these happen multiple times, which probably also creates a lot of churn that have no real effect. It might be worth investigating `notify_debouncer_mini` as a separate PR to make this less twitchy.

```
[00:00:30.041] (16bd9f000) INFO   File changed: "/Users/jonathan/Coding/gitu/src/file_watcher.rs" (Modify(Metadata(Any)))
[00:00:30.041] (16bd9f000) INFO   File changed: "/Users/jonathan/Coding/gitu/src/file_watcher.rs" (Modify(Data(Content)))
[00:00:30.042] (16bd9f000) INFO   File changed: "/Users/jonathan/Coding/gitu/src/file_watcher.rs~" (Create(File))
[00:00:30.042] (16bd9f000) INFO   File changed: "/Users/jonathan/Coding/gitu/src/file_watcher.rs~" (Remove(File))
[00:00:30.042] (16bd9f000) INFO   File changed: "/Users/jonathan/Coding/gitu/src/file_watcher.rs~" (Modify(Metadata(Any)))
[00:00:30.042] (16bd9f000) INFO   File changed: "/Users/jonathan/Coding/gitu/src/file_watcher.rs~" (Modify(Metadata(Ownership)))
[00:00:30.042] (16bd9f000) INFO   File changed: "/Users/jonathan/Coding/gitu/src/file_watcher.rs~" (Modify(Metadata(Extended)))
[00:00:30.043] (16bd9f000) INFO   File changed: "/Users/jonathan/Coding/gitu/src/file_watcher.rs~" (Modify(Data(Content)))```

> But let's leave the tests aside. I believe it'd be messy, since file update events would come in async. And it'd need to have some sort of sleep duration in hope of catching it.

Okay. I'll give it some thought, it would be nice to invest in those tests at some stage given the number of issues that have come up in this fairly small set of code.

File: src/file_watcher.rs [hunk 1]
```diff
@@ -1,39 +1,33 @@
...
         let pending_updates = Arc::new(AtomicBool::new(false));
         let pending_updates_w = pending_updates.clone();
 
-        let gitignore = GitignoreBuilder::new(path)
-            .add_line(None, super::LOG_FILE_NAME)
-            .map_err(Error::FileWatcherGitignore)?
-            .build()
-            .unwrap();
-
+        let repo = open_repo(repo_dir)?;
```
altsem, line 21: I recently merged a whole bunch of refactoring, where the creation of the file watcher was moved. Now it'd be real easy to just accept a `Repository` instead of a `Path` as argument. Then you'd don't need to call `open_repo` anymore!

It's created here! https://github.com/altsem/gitu/blob/e7b7719104adf9168b7fadf35b5899f2f425b43e/src/state.rs#L91

altsem: (I'll test the PR out again soon, but looks promising!)

altsem:
Gj on this! I feel relieved the file watcher is in a better state.
I'll merge it in!

altsem (approved):

PR #364, opened 2025-05-01
feat: wrap lines longer than the terminal width

### LINKED ISSUE #277: Text wrap git diff
When viewing diffs with long lines (e.g. Tailwind styles), the line is truncated by gitu making them unhelpful. An option to make the lines wrap would be really useful.

image

jonathanj:
Previously long lines would be truncated, now instead they are wrapped and a wrap indicator displayed at the front of each line, after the first one.

The wrap indicator and collapse indicators are configurable.

image

Fixes #277

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ pretty_assertions = "1.4.0"
 temp-dir = "0.1.13"
 criterion = "0.5.1"
 insta = "1.42.2"
-unicode-width = "0.2.0"
 
 [profile.release]
 strip = true
@@ -64,3 +63,4 @@ tree-sitter-elixir = "=0.1.1"
 regex = "1.11.1"
 strip-ansi-escapes = "0.2.1"
 unicode-segmentation = "1.12.0"
+unicode-width = "0.2.0"
--- a/src/config.rs
+++ b/src/config.rs
@@ -48,6 +48,8 @@ pub struct StyleConfig {
     pub selection_line: StyleConfigEntry,
     pub selection_bar: SymbolStyleConfigEntry,
     pub selection_area: StyleConfigEntry,
+    pub wrap_indicator: SymbolStyleConfigEntry,
+    pub collapsed_indicator: SymbolStyleConfigEntry,
 
     pub hash: StyleConfigEntry,
     pub branch: StyleConfigEntry,
--- a/src/default_config.toml
+++ b/src/default_config.toml
@@ -65,6 +65,8 @@ selection_line = { mods = "BOLD" }
 # You may want to set `selection_area.bg` to a nice background color.
 # Looks horrible with regular terminal colors, so is therefore not set.
 selection_area = {}
+wrap_indicator = { symbol = "↪", fg = "grey", mods = "DIM" }
+collapsed_indicator = { symbol = "…", fg = "grey", mods = "DIM" }
 
 hash = { fg = "yellow" }
 branch = { fg = "green" }
--- a/src/screen/mod.rs
+++ b/src/screen/mod.rs
@@ -384,6 +384,9 @@ impl Widget for &Screen {
             }
 
             line_index += height;
+            if line_index >= area.height {
+                break;
+            }
         }
     }
 }
--- a/src/screen/mod.rs
+++ b/src/screen/mod.rs
@@ -4,6 +4,8 @@ use crate::{config::Config, items::TargetData, Res};
 
 use super::Item;
 use std::{borrow::Cow, collections::HashSet, rc::Rc};
+use ratatui::widgets::{Paragraph, Wrap};
+use unicode_width::UnicodeWidthStr;
 
 pub(crate) mod log;
 pub(crate) mod show;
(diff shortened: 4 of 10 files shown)


jonathanj:
I have added tests and snapshots for the wrapping behaviour, to demonstrate some aspects of it like the cursor taking up multiple lines, etc. but I haven't updated all the existing snapshots because there is a potentially controversial change:

I opted to widen the "gutter" (where the cursor is displayed) to 2 columns, one for the cursor and one for the wrap indicator. From my perspective showing the wrap indicator is a much better user experience than showing nothing, because otherwise it feels like the cursor behaves inconsistently (since you can't always see what's a wrap without the indicator.)

I wanted to indent the wrapped lines by 1, so that I could make the extra gutter space only when needed, but `Paragraph` in ratatui [does not support this yet](https://github.com/ratatui/ratatui/issues/284). I also tried [a suggestion](https://github.com/ratatui/ratatui/issues/1798#issuecomment-2815992090) to use the `textwrap` crate, but that doesn't preserve styled Ratatui spans. There is also [this gist that implements custom ratatui wrapping](https://gist.github.com/sunshowers/457102aabd9db3bb23d61f2b4259bdd0), but that seemed a bit overkill. So widening the gutter by 1 seemed like the lesser evil.

The problem is that this would require me to update all the other snapshots, and I wanted to get your opinion on this before going off and bloating the PR with all of those changes.

jonathanj:
As far as the implementation goes, I don't know if this is a good approach. I thought that using ratatui's `Paragraph` to render this with wrapping would be fairly quick win (which it was), and the rest of the changes are mostly just making sure that `line_index` is incremented according to the additional screen lines needed for wrapping, and stretching out the cursor, etc.

`measure_paragraph_height` might seem a bit odd, but I couldn't find a way to have ratatui tell me how many lines will be required to wrap some text, and that's necessary to configure the render rect/area.

jonathanj:
There is currently at least 1 bug (and I'd be surprised if there aren't more), which I can't seem to track down:

Selecting an item that is wrapped will only highlight the first line but not the subsequent lines. I would have thought that `buf.set_style(line_area, &style.selection_line)` would work as expected, especially since `line_area` now has a height set.

jonathanj:
I realised that since there's now space at the front of the line, and the first line doesn't show a wrap indicator, it's possible to move the collapsed indicator there instead now. I didn't actually make this change, but here's what it could look like.

image

image

altsem:
Text wrapping is a great addition!

About adding an exta column to the gutter, I suppose it's possible.
An alternative idea I got was to replace the blue pipe on the left (selection indicator) with `⮓`. After all, I suppose you wouldn't be able to select a continuation of a line (ctrl+j, ctrl+k) anyway:
```
▌@@ -25,9 +25,9 @@ fn copy_hash(r: String) -> Option<Action> {
▌         match &mut state.clipboard {
▌             Some(cb) => {
▌                 cb.set_text(r.clone()).map_err(Error::Clipboard)?;
▌-                state.display_info("Commit hash copied to clipboar
↪d".to_owned());
▌+                state.display_info("Commit hash copied to clipboar
↪d");
▌             }
```

Or even doing both?
```
▌ @@ -25,9 +25,9 @@ fn copy_hash(r: String) -> Option<Action> {
▌          match &mut state.clipboard {
▌              Some(cb) => {
▌                  cb.set_text(r.clone()).map_err(Error::Clipboard)?;
▌ -                state.display_info("Commit hash copied to clipboar
 ↪d".to_owned());
▌ +                state.display_info("Commit hash copied to clipboar
 ↪d");
▌              }
```

One currently annoying thing though, if you double-click to select a whole word in a terminal, say you want to select and copy the commit hash:
```
▌da1d201 main yada yada
```
Then this would be selected: `▌da1d201`

#### Termwiz
I've been peeking at perhaps switching over and rewriting the rendering code to use Termwiz entirely instead. It's more low-level, but also feels more intuitive in that widgets can actually have a determined size (Ratatui seems to be all around deciding the size of containers first, and then the widget will have to make use of the size its given).

#### Navigation
I think there is the assumption in this module when it comes to navigation where it assumes an Item is always a single line. It might be that there's some weird behavior upon navigating if a lot of line-wrapping is in place.
Perhaps some preprocessing could be done to accommodate for this? There is already a notion of mapping visual lines to items like:
```
Line | Item | Content
6      6      ▌modified   src/state.rs
7      7      ▌@@ -87,15 +87,7 @@ impl State {…
8      24     ▌@@ -108,9 +100,27 @@ impl State {
9      25     ▌             current_cmd_log: CmdLog::new(),
```

Found here:
https://github.com/altsem/gitu/blob/6b9e124d0c399f93238770d39d1cee1f17aef751/src/screen/mod.rs#L235-L256

jonathanj:
> Or even doing both?

Leaving the cursor/selection out for a continuation is a good idea.

> One currently annoying thing though, if you double-click to select a whole word in a terminal, say you want to select and copy the commit hash:
> 
> Then this would be selected: `▌da1d201`

Adding the extra gutter column seems to improve this, at least in my terminal.

> I think there is the assumption in this module when it comes to navigation where it assumes an Item is always a single line. It might be that there's some weird behavior upon navigating if a lot of line-wrapping is in place. Perhaps some preprocessing could be done to accommodate for this? There is already a notion of mapping visual lines to items like:

The navigation seems to work okay, at least as far as I could tell. What could I try to produce a mismatched line issue? I figured that since the line wrapping is happening only at the render point, that what happens in effect is that the virtual lines (for items) are still single lines, but at the point they become characters in the terminal then they end up being wrapped, although the internal navigation logic doesn't know (or care) that the renderer does that.

altsem:
One example is when navigating below the visible view, the view should scroll along with the cursor. But if enough lines are wrapped, then it would be delayed.

There is logic in the screen module that makes assumptions about navigation from the size of the view. The `size` property of `Screen`.

PR #367, opened 2025-05-14
feat: add delete option to branch menu

kamilwaz:
This PR adds support for deleting branches from the branch menu. If a branch hasn't been merged into upstream or HEAD, the operation must be confirmed (similar to how to `git branch -d ...` works).

### DIFF
--- a/src/ops/branch.rs
+++ b/src/ops/branch.rs
@@ -1,5 +1,7 @@
 use super::{selected_rev, Action, OpTrait};
 use crate::{
+    error::Error,
+    git::{get_current_branch_name, is_branch_merged},
     items::TargetData,
     menu::arg::Arg,
     state::{PromptParams, State},
@@ -37,9 +39,7 @@ impl OpTrait for Checkout {
 
 fn checkout(state: &mut State, term: &mut Term, rev: &str) -> Res<()> {
     let mut cmd = Command::new("git");
-    cmd.args(["checkout"]);
-    cmd.args(state.pending_menu.as_ref().unwrap().args());
-    cmd.arg(rev);
+    cmd.args(["checkout", rev]);
 
     state.close_menu();
     state.run_cmd(term, &[], cmd)?;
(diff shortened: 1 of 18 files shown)


src/ops/branch.rs
```diff
@@ -80,3 +80,64 @@ fn checkout_new_branch_prompt_update(
...
+    let mut cmd = Command::new("git");
+    cmd.args(["branch", "-d"]);
+
+    if !is_branch_merged(&state.repo, branch_name).unwrap_or(false) {
+        match state.confirm(term, "Branch is not fully merged. Really delete? (y or n)") {
+            Ok(()) => {
+                cmd.arg("-f");
+            }
+            Err(err) => {
+                state.close_menu();
```
kamilwaz, line 132: This line prevents the menu from getting stuck. Without it, the menu is hidden, but you can't navigate unless you press `q` or `<esc>`.

It seems to be a global issue. For example, pressing `Pp<enter>` causes the same bug. If that gets fixed, we could simplify this part like this:

```rust
    if !is_branch_merged(&state.repo, branch_name).unwrap_or(false) {
        state.confirm(term, "Branch is not fully merged. Really delete? (y or n)")?;
        cmd.arg("-f");
    }
```

altsem: :thinking: this might've been a regression from all my fiddling around.
I think it makes sense to go with the simpler solution, and let's acknowledge the problem in this PR and fix it in another one before releasing the next version.
I can try have a look.

I butchered your sentence and made a new issue :D
https://github.com/altsem/gitu/issues/368

src/tests/branch.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+#[test]
+fn delete_branch_selected() {
+    snapshot!(setup(), "Yjjbd<enter>");
+}
+
+#[test]
+fn delete_branch_input() {
+    snapshot!(setup(), "bcmergedll");
```
altsem, line 39: Looks like a copypasta, this is creating a branch.
And missing the <enter> at the end if you intended to actually perform the deletion.

In other cases I've gone so far to even have test snapshots of prompts, which is fine too:

Something like:
"delete_branch_by_input_prompt": `bdmerged`
"delete_branch_by_input": `bdmerged<enter>`

src/tests/branch.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+#[test]
+fn delete_branch_empty() {
+    snapshot!(setup(), "bd<enter>");
+}
+
+#[test]
+fn delete_unmerged_branch() {
+    snapshot!(setup(), "Yjjjbd<enter>nbd<enter>y");
```
altsem, line 49: I'd be happy with just going by name here instead, `bdunmerged<enter>...`

altsem (changes requested):
`feat: add delete branch` goes straight to the changelog (we've got changelog generation set up).
So perhaps something just a little more descriptive to the users would be nice.

There is actually another place where this same "prompt y or n to delete branch" should be happening, and that is in the "discard" action. Not a requirement for this PR, include it if you wish, else we'll fix it later.

relates to: https://github.com/altsem/gitu/pull/350

src/tests/branch.rs
kamilwaz: I thought about adding support for multiple snapshots per test, but didn’t want to complicate this PR. For now, I'll just replace the character and add `<enter>` at the end.

src/tests/snapshots/gitu__tests__branch__delete_branch_forcefully.snap
altsem, line 1: This snapshot seems stale

altsem (approved):
ill merge it, gj!

PR #377, opened 2025-05-29
feat: discard line by line (+ configure when to confirm discard)

satoqz:
Adds discarding unstaged changes line by line.

By default you'd have to confirm each line that you unstage separately when unstaging several lines at once. As that might get a little tedious, I've added a config option `confirm_discard` that lets you choose the granularity of when to ask for confirmation:

```toml
confirm_discard = "line"  # asks on files/hunks/lines
confirm_discard = "hunk"  # asks on files/hunks
confirm_discard = "file"  # asks on files
confirm_discard = "never" # never asks, scary
```

The default is `"line"`, so you are asked to confirm no matter what you discard, which is backwards-compatible and probably the best for new users.

Curious what you think!

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -23,6 +23,7 @@ pub struct GeneralConfig {
     pub always_show_help: BoolConfigEntry,
     pub confirm_quit: BoolConfigEntry,
     pub refresh_on_file_change: BoolConfigEntry,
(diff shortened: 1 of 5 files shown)


altsem (approved):
This is awesome! I'll merge it in. :)

PR #386, opened 2025-06-12
fix: report key bindings without "<>"

### LINKED ISSUE #256: Config parsing shouldn't fail silently.
Spawned from: https://github.com/altsem/gitu/issues/255

loading a config with invalid bindings fails silently:
```toml
[bindings]
root.quit = ["ctrl+g", "<esc>"]
```

cjordan:
Fixes #256

This commit does a few things. Primarily, it validates that there are no key bindings like "ctrl+g" (these should be "<ctrl+g>", i.e. surrounded by angle brackets).

To ensure that the bindings are correct for the rest of the code, the `Config` struct has been split into something that `figment` recognises and another that's actually used.

Finally, if the whole program has an error, the `Display` representation is printed, not the `Debug` one; this makes the error messages easier for users to read.

It would be better to detect errors in the `nom` code, but I lack experience with this and get a headache every single time :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -796,6 +796,7 @@ dependencies = [
  "similar",
  "simple-logging",
  "strip-ansi-escapes",
+ "strum",
  "temp-dir",
  "termwiz",
  "toml",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,3 +63,4 @@ strip-ansi-escapes = "0.2.1"
 unicode-segmentation = "1.12.0"
 termwiz = "0.22.0"
 ratatui = { version = "0.29.0", features = ["termwiz", "serde", "underline-color"], default-features = false }
+strum = { version = "0.26.3", features = ["strum_macros"] }
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -3,29 +3,86 @@ use std::collections::BTreeMap;
 use termwiz::input::{KeyCode, Modifiers};
 
 use crate::{
+    error::Error,
     key_parser,
     menu::{Menu, PendingMenu},
     ops::Op,
 };
 
+// TODO: Make the nom parsing code return an error instead.
+const SPECIAL_KEYS: [&str; 19] = [
+    "backspace",
+    "enter",
+    "left",
+    "right",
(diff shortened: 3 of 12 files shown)


cjordan:
Here's an example of the error output:
```
Errors while parsing key bindings:
- root.move_parent_section = alt+left
- fetch_menu.quit = esc

Special keys like "ctrl" require key bindings to be surrounded, e.g. "<ctrl+g>"
```

I get this from changing these default bindings:
```
# root.move_parent_section = ["<alt+h>", "<alt+left>"]
root.move_parent_section = ["<alt+h>", "alt+left"]
# fetch_menu.quit = ["q", "<esc>"]
fetch_menu.quit = ["q", "esc"]
```

cjordan:
And, hi! Thanks for your work on this project.

I appreciate that my changes in this PR may not be to your taste. Please let me know if you'd like me to tweak it in some way.

There were quite a few ways to approach this but I tried to keep the code changes minimal.

altsem:
Hi! Thanks for looking into this, it's been a pain for some time.
I do think you have a point in trying to fix the problem in the parser, and have it output errors would be a nicer way to go. I think i realized something, posted it back in https://github.com/altsem/gitu/issues/256. See if that makes sense :)

altsem:
@cjordan Exactly what I thought! If the error message is clear enough, we may just release a new version and have people figure it out from that error.

Let's hold with this PR, and see where the other one turns out.

cjordan:
Hi again, I've attempted to do what we've discussed (see my newest commit) but I'm having issues with the tests. One of them even opens a prompt in my terminal to make a a git commit :) Not sure how to work around this... I'd appreciate some help!

altsem:
Hey @cjordan! Im without a computer until mid August. I'll try have a look at the code some time, but it's harder for me to help atm.

cjordan:
Hi, me again. I've made a new PR: https://github.com/altsem/gitu/pull/407

I appreciate that this is pretty big (unfortunately), and probably includes some opinionated choices. I'm happy to edit this as necessary, but hopefully this is looking good. Cheers!

PR #390, opened 2025-06-17
feat: ability to delete a remote

pritchett:
Adds the ability to remove a remote. Used `k` as that is what magit uses.

### DIFF
--- a/src/default_config.toml
+++ b/src/default_config.toml
@@ -172,6 +172,7 @@ rebase_menu.quit = ["q", "<esc>"]
 
 root.remote_menu=["M"]
 remote_menu.add_remote=["a"]
+remote_menu.remove_remote=["k"]
 remote_menu.rename_remote=["r"]
 remote_menu.quit = ["q", "<esc>"]
 
--- a/src/default_config.toml
+++ b/src/default_config.toml
@@ -172,7 +172,7 @@ rebase_menu.quit = ["q", "<esc>"]
 
 root.remote_menu=["M"]
 remote_menu.add_remote=["a"]
-remote_menu.remove_remote=["k"]
+remote_menu.remove_remote=["K"]
 remote_menu.rename_remote=["r"]
 remote_menu.quit = ["q", "<esc>"]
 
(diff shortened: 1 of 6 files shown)


src/default_config.toml
```diff
@@ -172,6 +172,7 @@ rebase_menu.quit = ["q", "<esc>"]
 
 root.remote_menu=["M"]
 remote_menu.add_remote=["a"]
+remote_menu.remove_remote=["k"]
```
altsem, line 175: We bound the new "delete branch" to "K" to be consistent with discarding stuff I believe.
I think this should have the same binding, it'd be nice for it to be consistent across the board.

src/ops/remote.rs
```diff
@@ -74,6 +74,44 @@ impl OpTrait for RenameRemote {
...
+            |state: &mut crate::state::State, term: &mut Term| {
+                let remote_name = state.prompt(
+                    term,
+                    &PromptParams {
+                        prompt: "Delete remote",
+                        ..Default::default()
+                    },
+                )?;
+
+                remove_remote(state, term, &remote_name)?;
```
altsem, line 90: There's a `state.confirm` function used elsewhere when removing stuff, should that be here as well?

altsem (changes requested):

src/default_config.toml
pritchett: That's been changed

src/tests/snapshots/gitu__tests__remote__remove_remote.snap
```diff
@@ -0,0 +1,25 @@
...
+                                                                                |
+                                                                                |
+                                                                                |
+                                                                                |
+                                                                                |
+                                                                                |
+                                                                                |
+────────────────────────────────────────────────────────────────────────────────|
+? Rename to: ›                                                                  |
+styles_hash: 2723afb20d7e3ef0
```
altsem, line 25: I think the test just needs updating, it must be read `M`, skip `k`, skip `o` and then pick up `r` as in "rename".

PR #391, opened 2025-06-17
additional rendering work

mWalrus:
Creating a draft PR to centralize conversation.

What has changed:
- `Item` no longer has field `display`
- `TargetData` has been extended to support all different variations required for rendering
- `TargetData` and its related types have moved into `src/target_data.rs`
- Rendering of `Item`s is now done when needed
- `TargetData::to_line` is used to construct `Line`s for rendering items
- Improve output string for branch status: "... commit" -> "... commit(s)"

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -37,8 +37,8 @@ use crate::prompt::PromptData;
 use crate::prompt::Status;
 use crate::screen;
 use crate::screen::Screen;
-use crate::target_data::RefKind;
-use crate::target_data::TargetData;
+use crate::item_data::RefKind;
+use crate::item_data::ItemData;
 use crate::term;
 use crate::term::Term;
 use crate::ui;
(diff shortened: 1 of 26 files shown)


altsem:
@mWalrus That other branch is pretty broken. Where as this feels like a nice isolated change.
Sorry for the mess, but I think we should base this branch off of master.

I have broken out some refactoring changes from feat/rendering-overhaul already and merged them to master.

I reckon you should be able to cherry-pick your commits one-by-one on top of master (perhaps there's a better way?) with minimal conflicts.

mWalrus:
@altsem alright! I'll look into doing that.

Also, I think I've figured out why gitu runs slow sometimes and I'm pretty sure it has to do with exactly what I'm working on. It's stuck pre-emptively "rendering" huge diffs to store in `Item.display` on startup.

Good news is that my changes should (hopefully) resolve this.

PR #276, opened 2024-10-27
feat(config): Collapse screen headers via e.g. `general.collapsed_sec…

altsem:
…tions = ["recent_commits"]`

Hide headers on the status screen e.g.
```
collapsed_sections = ["untracked", "recent_commits", "branch_status"]
```

@stackmystack I did some changes:
- added some consistency to the internal IDs
- changed the commit message (the changelog is auto-generated from commits)
- updated the `default_config.toml`

I'll go ahead and merge this!

PR #392, opened 2025-06-18
Feat/just in time rendering

mWalrus:
Cherry picked commits from #391.

# Todo
- Status section items are not being rendered
- Stash list section items are not being rendered
- Log section items are not being rendered

Above are not working because they are not being assigned `ItemData`.

Supersedes #391.

### DIFF
--- a/src/highlight.rs
+++ b/src/highlight.rs
@@ -14,19 +14,26 @@ use std::path::Path;
 use std::rc::Rc;
 use unicode_segmentation::UnicodeSegmentation;
 
-type LineHighlights<'a> = (&'a str, Vec<(Range<usize>, Style)>);
+type LineHighlights<'a> = Vec<(Range<usize>, Style)>;
 
 pub(crate) fn highlight_hunk_lines<'a>(
     config: &'a Config,
     diff: &'a Rc<Diff>,
     file_i: usize,
     hunk_i: usize,
-) -> impl Iterator<Item = LineHighlights<'a>> + 'a {
-    let old_path = &diff.text[diff.file_diffs[file_i].header.old_file.clone()];
-    let new_path = &diff.text[diff.file_diffs[file_i].header.new_file.clone()];
+) -> impl Iterator<Item = Vec<(Range<usize>, Style)>> + 'a {
+    let file_diff = &diff.file_diffs[file_i];
 
-    let hunk = &diff.file_diffs[file_i].hunks[hunk_i];
-    let hunk_content = &diff.text[hunk.content.range.clone()];
+    let old_file_range = file_diff.header.old_file.clone();
+    let new_file_range = file_diff.header.new_file.clone();
+
+    let old_path = &diff.text[old_file_range];
+    let new_path = &diff.text[new_file_range];
+
+    let hunk = &file_diff.hunks[hunk_i];
+    let hunk_range = hunk.content.range.clone();
+    let hunk_content = &diff.text[hunk_range];
+
     let old_mask = diff.mask_old_hunk(file_i, hunk_i);
     let new_mask = diff.mask_new_hunk(file_i, hunk_i);
 
@@ -46,10 +53,7 @@ pub(crate) fn highlight_hunk_lines<'a>(
     hunk_content
         .split_inclusive('\n')
         .scan_byte_ranges()
-        .map(move |(line_range, line)| {
-            let highlights = collect_line_highlights(&mut highlights_iter, &line_range);
-            (line, highlights)
-        })
+        .map(move |(line_range, _)| collect_line_highlights(&mut highlights_iter, &line_range))
 }
 
 pub(crate) fn iter_diff_highlights<'a>(
--- a/src/highlight.rs
+++ b/src/highlight.rs
@@ -154,10 +154,23 @@ trait ScanByteRanges<T> {
 
 impl<'a, I: Iterator<Item = &'a str>> ScanByteRanges<&'a str> for I {
     fn scan_byte_ranges(self) -> impl Iterator<Item = (Range<usize>, &'a str)> {
-        self.scan(0..0, |prev_line_range, line| {
-            let line_range = prev_line_range.end..(prev_line_range.end + line.len());
-            *prev_line_range = line_range.clone();
-            Some((line_range, line))
+        self.scan(0..0, |previous_line, current_line| {
+            let line_start = previous_line.end;
+            let actual_line_length = current_line.len();
+
+            let line_end = if current_line.ends_with("\r\n") {
+                // subtract \r length + \n length from scanned line to "trim" the line
+                actual_line_length.saturating_sub(2)
+            } else {
+                actual_line_length
+            };
+
+            let valid_line_range = line_start..(line_start + line_end);
+
+            let scanned_line_range = line_start..(line_start + actual_line_length);
+            *previous_line = scanned_line_range.clone();
+
+            Some((valid_line_range, current_line))
         })
     }
 }
--- a/src/item_data.rs
+++ b/src/item_data.rs
@@ -5,7 +5,7 @@ use ratatui::text::{Line, Span};
 use crate::{config::Config, git::diff::Diff, gitu_diff::Status};
 
 #[derive(Clone, Debug)]
-pub(crate) enum TargetData {
+pub(crate) enum ItemData {
     Empty,
     AllUnstaged,
     AllStaged,
@@ -63,18 +63,16 @@ pub(crate) enum SectionHeader {
     Revert(String),
 }
 
-impl TargetData {
+impl ItemData {
     pub fn to_line<'a>(&'a self, config: Rc<Config>) -> Line<'a> {
         match self {
-            TargetData::Empty => Line::raw(""),
-            TargetData::AllUnstaged => {
-                Line::styled("Unstaged changes", &config.style.section_header)
-            }
-            TargetData::AllStaged => Line::styled("Staged changes", &config.style.section_header),
-            TargetData::AllUntracked(_) => {
+            ItemData::Empty => Line::raw(""),
+            ItemData::AllUnstaged => Line::styled("Unstaged changes", &config.style.section_header),
+            ItemData::AllStaged => Line::styled("Staged changes", &config.style.section_header),
+            ItemData::AllUntracked(_) => {
                 Line::styled("Untracked files", &config.style.section_header)
             }
-            TargetData::Reference(ref_kind) => {
+            ItemData::Reference(ref_kind) => {
                 let (reference, style) = match ref_kind {
                     RefKind::Tag(tag) => (tag, &config.style.tag),
                     RefKind::Branch(branch) => (branch, &config.style.branch),
@@ -83,7 +81,7 @@ impl TargetData {
                 // TODO create prefix
                 Line::styled(reference, style)
             }
-            TargetData::Commit {
+            ItemData::Commit {
                 short_id,
                 associated_references,
                 summary,
@@ -111,10 +109,8 @@ impl TargetData {
 
                 Line::from(spans)
             }
-            TargetData::File(path) => {
-                Line::styled(path.to_string_lossy(), &config.style.file_header)
-            }
-            TargetData::Delta { diff, file_i } => {
+            ItemData::File(path) => Line::styled(path.to_string_lossy(), &config.style.file_header),
+            ItemData::Delta { diff, file_i } => {
                 let file_diff = &diff.file_diffs[*file_i];
 
                 let content = format!(
@@ -132,7 +128,7 @@ impl TargetData {
 
                 Line::styled(content, &config.style.file_header)
             }
-            TargetData::Hunk {
+            ItemData::Hunk {
                 diff,
                 file_i,
                 hunk_i,
@@ -144,7 +140,7 @@ impl TargetData {
 
                 Line::styled(content, &config.style.hunk_header)
             }
-            TargetData::HunkLine {
+            ItemData::HunkLine {
                 diff,
                 file_i,
                 hunk_i,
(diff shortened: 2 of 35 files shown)


src/items.rs
```diff
@@ -1,68 +1,37 @@
...
+    // preferably we can store text and styling separately like in highlight.rs: `(Range<usize>, Style)`
+    // and only apply them when rendering
     pub(crate) section: bool,
     pub(crate) default_collapsed: bool,
     pub(crate) depth: usize,
     pub(crate) unselectable: bool,
-    pub(crate) target_data: Option<TargetData>,
+    // TODO rename? maybe item_data: Option<ItemData>
+    // TODO does this have to be optional anymore?
+    pub(crate) data: Option<ItemData>,
```
altsem, line 32: I like `data`! It's under `Item`, and `ItemData` seems fine.
I thought of this too, I think it wont't need to be Optional.

src/items.rs
```diff
@@ -1,68 +1,37 @@
...
 
 pub type ItemId = u64;
 
 #[derive(Default, Clone, Debug)]
 pub(crate) struct Item {
     pub(crate) id: ItemId,
-    pub(crate) display: Line<'static>,
+    // TODO We'll want to move away from this `Line` struct
+    // preferably we can store text and styling separately like in highlight.rs: `(Range<usize>, Style)`
+    // and only apply them when rendering
```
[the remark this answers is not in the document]
mWalrus: removed in https://github.com/altsem/gitu/pull/392/commits/e2bfa944848f12aec846d1310329311f5a106b33.

mWalrus:
Crashing somewhere here at the moment:
https://github.com/altsem/gitu/blob/1d1a7a77f40f5d508b90a8dd034e64873aaf264e/src/screen/mod.rs#L330-L373

ERROR: `index outside of buffer: the area is Rect { x: 0, y: 0, width: 213, height: 49 } but index is (0, 49)`

Reading errors is really tricky currently, maybe I should implement a panic hook that logs the stacktrace to a file somewhere in a separate PR. What do you think about that @altsem ?

mWalrus:
I've been hacking away some more and the current problem is rendering `HunkLine`s.
In master we construct each display when the `HunkLine` is created after `highlight::highlight_hunk_lines` has done its thing.

We have since moved over highlighting to the render step but the problem is now that we need to rework `src/highlight.rs` to support line by line highlighting. This is so that we only have to highlight, style and render the hunk lines that are visible instead of the entire diff.

I skimmed through `src/highlight.rs` a little to see what could be done but I haven't come up with anything yet.

`HunkLine`s are currently rendered unhighlighted.

mWalrus:
@altsem,

# Syntax highlighting
Had never heard of cached before but it looked pretty fitting so I tried my hand at implementing some caching for syntax highlighting and it seems to be working! I'm currently failing a bunch of tests but most of the original functionality seems to be restored. Will take a closer look at tests next.

One thing I've also noticed is that even if we don't highlight the diffs immediately anymore, we parse and prepare diffs for highlighting even if they are not being rendered on the screen. Not sure if this means that they'll get processed for syntax highlighting anyways because of how `Item`s are being taken care of when drawing a frame but I fear that this might be the case.

# Panic hook
I've moved discussion to https://github.com/altsem/gitu/issues/393 :)

altsem:
Awesome! I do actually believe highlight_hunk was the expensive function you encountered before. It performs all highlighting, diff and syntax, and merges them.

The entire diff is still interpreted and stored in `Diff` before highlighting, but this shouldn't be all too slow.

Now if a line is requested, it should go generate the hunk with all styles on the fly 💪

mWalrus:
Still trying to pass tests but am stuck on syntax highlighting and my brain is breaking! :laughing:
Everything looks to be correct when viewing hunk diffs in the tool:
image
But a bunch of diff highlight involving tests fail:
image

mWalrus:
omg @altsem... I realized why tests are failing :laughing: My current implementation of `highlight_hunk`'s cache-key generation was causing collisions between tests!

```rust
#[cached(
    ty = "SizedCache<String, HunkHighlights>",
    create = "{ SizedCache::with_size(100) }",
    convert = r#"{ format!("file{}::hunk{}", file_index, hunk_index)}"# // <--- THIS is the culprit
)]
pub(crate) fn highlight_hunk<'a>(
    config: &'_ Config,
    diff: &'_ Rc<Diff>,
    file_index: usize,
    hunk_index: usize,
) -> HunkHighlights {
    // ...
}
```

I'll see if I can create a better cache-key generation step. Maybe a hash of some sort. Please let me know if you have any ideas as well!

altsem:
@mWalrus i made some hashes for other purposes the other day here :) https://github.com/altsem/gitu/blob/master/src/items.rs#L126

It's needed to populate Item.id.
Which is used to identify collapsed/expanded sections.

Perhaps there's a way to compute it just once? Else it might not be too expensive to call again

src/item_data.rs
```diff
@@ -0,0 +1,216 @@
...
+    UpstreamGone(String),
+    Rebase(String, String),
+    Merge(String),
+    Revert(String),
+    Stashes,
+    RecentCommits,
+    Commit(String),
+}
+
+impl ItemData {
```
altsem, line 79: I thought if this before, but now that you do need the `id` of an item - it feels even more relevant.
What if:
```suggestion
impl Item {
```

Then you could skip the item_id argument! :)

src/highlight.rs
```diff
@@ -14,42 +15,65 @@ use std::path::Path;
...
-    let new_path = &diff.text[diff.file_diffs[file_i].header.new_file.clone()];
-
-    let hunk = &diff.file_diffs[file_i].hunks[hunk_i];
-    let hunk_content = &diff.text[hunk.content.range.clone()];
-    let old_mask = diff.mask_old_hunk(file_i, hunk_i);
-    let new_mask = diff.mask_new_hunk(file_i, hunk_i);
+#[cached(
+    ty = "SizedCache<String, HunkHighlights>",
+    create = "{ SizedCache::with_size(200) }",
+    convert = r#"{ format!("{_hunk_hash}") }"#
```
altsem, line 21:
```suggestion
    ty = "SizedCache<u64, HunkHighlights>",
    create = "{ SizedCache::with_size(200) }",
    convert = r#"{ _hunk_hash }"#
```

I suspect something like this might work now

altsem (commented):
Checking in on the progress and dumped some thoughts.
I think this is coming along really well. Feels like pieces are falling into place.

Will try poke around some more soon again

src/highlight.rs
```diff
@@ -14,42 +15,65 @@ use std::path::Path;
...
+    HunkHighlights { spans }
+}
+
+#[derive(Clone)]
+pub struct HunkHighlights {
+    spans: Vec<Vec<(Range<usize>, Style)>>,
+}
+
+impl HunkHighlights {
+    pub fn get_hunk_line(&self, line_range: usize) -> &[(Range<usize>, Style)] {
```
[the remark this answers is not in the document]
mWalrus: This method returns all the highlight ranges and styles for a given line. But, yeah, it is in need of some clean up for sure! Made an attempt here https://github.com/altsem/gitu/pull/392/commits/8abd9d7a95cd276b7e2c6d371f7ca2373673d969.
Let me know if you have any other ideas :)

src/highlight.rs
mWalrus: You're correct! I completely missed this :sweat_smile: Done in https://github.com/altsem/gitu/pull/392/commits/39aa51df0d65569991eadee735bf4b4d596fab0a.

src/ops/branch.rs
```diff
@@ -79,9 +79,14 @@ fn checkout_new_branch_prompt_update(app: &mut App, term: &mut Term, branch_name
...
 impl OpTrait for Delete {
-    fn get_action(&self, target: Option<&TargetData>) -> Option<Action> {
+    fn get_action(&self, target: &ItemData) -> Option<Action> {
         let default = match target {
-            Some(TargetData::Branch(b)) => Some(b.clone()),
+            ItemData::Reference { kind, .. } => match kind {
+                RefKind::Tag(tag) => Some(tag.clone()),
+                RefKind::Branch(branch) => Some(branch.clone()),
+                // FIXME is this correct?
+                RefKind::Remote(_) => None,
```
altsem, line 88: This action will actually run `git branch -d` as seen down below. Does that work with tags? I guess not for remotes.

What do you think of leaving it similar to the previous one just (ignore Tag/Remote, just deal with Branch)?

src/ops/discard.rs
```diff
@@ -6,40 +6,48 @@ use crate::{
...
-            Some(TargetData::Delta { diff, file_i }) => match diff.file_diffs[file_i].header.status
-            {
+    fn get_action(&self, target: &ItemData) -> Option<Action> {
+        let action = match target {
+            ItemData::Reference { kind, .. } => {
+                let branch = match kind {
+                    RefKind::Tag(tag) => tag,
+                    RefKind::Branch(branch) => branch,
+                    // FIXME is this correct?
+                    RefKind::Remote(_) => unreachable!("cannot discard remotes"),
```
altsem, line 22: You could do a `return None` for now, should consider this in https://github.com/altsem/gitu/pull/390

src/ops/branch.rs
altsem: Oh wait, I saw the code in `discard.rs`, so it works with tags too? Then we could leave it.

src/screen/status.rs
```diff
@@ -177,92 +150,62 @@ fn branch_status_items(config: &Config, repo: &Repository) -> Res<Vec<Item>> {
...
             Item {
-                id: hash(snake_case_header),
-                display: Line::from(vec![
-                    Span::styled(
-                        capitalize(&snake_case_header.replace("_", " ")),
-                        &style.section_header,
-                    ),
-                    format!(" ({})", diff.file_diffs.len()).into(),
-                ]),
+                id: hash(section),
```
altsem, line 182: This might break what was implemented back in https://github.com/altsem/gitu/pull/276.
As seen here in the default config too:
https://github.com/altsem/gitu/blob/58cb7870757a4b8d430b96a43b54e1a666a625ff/src/default_config.toml#L8-L10

altsem (changes requested):
Found just one potential issue with the collapsed section config.
I'm curious how this will perform, if it solves that nasty lag in larger projects.

altsem:
Oh, and, I reviewed everything and think we're basically good to go with the aforementioned things fixed!

The changelog generation is currently based on git commits, but I'm considering trashing it and doing it manually at some point...

Or if we squash everything in here. :thinking: 

```
> git cliff --unreleased
# Changelog

All notable changes to this project will be documented in this file.

## [unreleased]

### 🚀 Features

- Cache syntax highlighting for hunks
- `ItemData::to_line` -> `ItemData::to_lines`
- Implement commit rendering
- JIT rendering
- Ability to rename a remote
- Switch terminal backend from Crossterm to Termwiz

### 🐛 Bug Fixes

- Don't format!() hunk highlight cache-keys
- Use already computed hunk hash for caching
- Return commits from `selected_rev` again
- Restore ref styling
- Replace tab-characters with 4 spaces again
- Generate more collision resistant cache-key
- Restore stash item styling
- `Diff::hunk_content` to unify getting hunk content
- Restore creation of `RefKind::Remote` items
- Store visual reference prefix to apply later
- Unify hunk line range scanning into one fn
- Typo in stash string
- Restore status section headers
- Restore commit view header
- `ItemData::Raw` holds an owned String
- `ItemData::to_line` now produces a single line
- (mostly) restore hunk line rendering
- Correctly highlight hunks again
- Removing very hacky logging that made it thru
- Make `Item.data` is no longer optional
- Adding some missing section headers
- Construct lines again
- Restore replace of TAB with 4 spaces
- Typo in branch status text

### ⚡ Performance

- Avoid some allocations
- More efficiently keep track of changes between updates

<!-- generated by git-cliff -->
```

mWalrus:
@altsem,
We can squash the commits to make it easier. We can reword it so it makes more sense as well! :blush:

src/item_data.rs
```diff
@@ -0,0 +1,219 @@
...
+    UpstreamGone(String),
+    Rebase(String, String),
+    Merge(String),
+    Revert(String),
+    Stashes,
+    RecentCommits,
+    Commit(String),
+}
+
+impl Item {
```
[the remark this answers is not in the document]
mWalrus: done in https://github.com/altsem/gitu/pull/392/commits/86752b054c13f075141cf72ec8c2774834b40ee7.

src/ops/discard.rs
mWalrus: see https://github.com/altsem/gitu/pull/392/commits/d05947a7d6438cb082366657fc711698d1853a59

src/screen/status.rs
mWalrus: I made an attempt in https://github.com/altsem/gitu/pull/392/commits/f732d3effb3c372b3aa2424a115c5429269b7f92.

altsem:
I'll merge this, let it sit in master for a few days, perhaps bundle some more stuff in, then include it in a release soon :) gj!

jonathanj:
This is a great improvement, I recently had a case where highlighting (a brutal JSON file) caused a lot of slowdown, so thanks!

> The changelog generation is currently based on git commits, but I'm considering trashing it and doing it manually at some point... https://github.com/altsem/gitu/blob/master/docs/dev-tooling.md
> 
> Or if we squash everything in here. 🤔

I noticed this, and thought I'd point out that you could configure git-cliff to [filter based on merge commit](https://git-cliff.org/docs/tips-and-tricks#filter-merge-commits), so if you discard non-merge commits then as long as your merge commit is conventional, it would be a "one PR equals one changelog item" workflow.

PR #396, opened 2025-06-27
feat: commit extend

digyx:
Magit includes a commit extend op which allows you to amend a commit using the same commit message.

https://github.com/magit/magit/blob/a4f73fb2fb55f7644a80b4442379ef43840ec5e9/lisp/magit-commit.el#L206

This PR implements the commit extend operation without the override date.


altsem (approved):
I played around with this a bit, and I stumbled upon some prompts when using it in the original Magit. They don't seem crucial to me.

```
Nothing staged.  Stage and commit all unstaged changes? (y or n)
```
```
This commit has already been published to origin/master.
Do you really want to modify it? (y or n)
```

PR #407, opened 2025-08-17
fix: don't require <> in config key parsing

cjordan:
This commit changes the `nom` key parsing code such that key bindings in a user's config do not surround special keys (e.g. "ctrl") in angle brackets. Unfortunately this means many tests need to be changes.

Also, any errors encountered when parsing a user's binding config are immediately reported to the user.

### DIFF
--- a/src/key_parser.rs
+++ b/src/key_parser.rs
@@ -127,7 +127,7 @@ fn parse_normal_key(input: &str) -> IResult<&str, KeyCode> {
 }
 
 fn parse_char_key(input: &str) -> IResult<&str, KeyCode> {
-    map(anychar, |c| KeyCode::Char(c))(input)
+    map(anychar, KeyCode::Char)(input)
 }
 
 fn parse_special_key(input: &str) -> IResult<&str, KeyCode> {
--- a/src/app.rs
+++ b/src/app.rs
@@ -86,7 +86,7 @@ impl App {
         let pending_menu = root_menu(&config).map(PendingMenu::init);
 
         let clipboard = Clipboard::new()
-            .inspect_err(|e| log::warn!("Couldn't initialize clipboard: {}", e))
+            .inspect_err(|e| log::warn!("Couldn't initialize clipboard: {e}"))
             .ok();
 
         let mut app = Self {
@@ -369,7 +369,7 @@ impl App {
             return Ok(());
         };
 
-        log::debug!("pending cmd finished with {:?}", status);
+        log::debug!("pending cmd finished with {status:?}");
 
         let result = write_child_output_to_log(log_rwlock, child, status);
         self.state.pending_cmd = None;
--- a/src/cmd_log.rs
+++ b/src/cmd_log.rs
@@ -91,14 +91,11 @@ pub(crate) fn format_log_entry<'a>(
         }))
         .collect::<Vec<_>>(),
         CmdLogEntry::Error(err) => {
-            vec![Line::styled(
-                format!("! {}", err),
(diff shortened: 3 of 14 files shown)


altsem:
Started reviewing this now. Think it looks pretty good so far.
I guess there'll be a discrepancy between the `-k` flag and bindings in the config file.
Perhaps it would be nice if they were consistent, but at least it's documented in `--help`.

src/key_parser.rs
```diff
@@ -1,77 +1,171 @@
...
+            map(tag("pageup"), |_| KeyCode::PageUp),
+            map(tag("pagedown"), |_| KeyCode::PageDown),
+            map(tag("tab"), |_| KeyCode::Tab),
+            map(tag("delete"), |_| KeyCode::Delete),
+            map(tag("insert"), |_| KeyCode::Insert),
+            map(tag("esc"), |_| KeyCode::Escape),
+            map(tag("capslock"), |_| KeyCode::CapsLock),
+        )),
+        // TODO: I got a compile error when trying to put the F keys above.
+        // Can this be cleaned up?
```
altsem, line 153: `alt` has a limit of tuples with 21 entries at most.
Let's just keep this TODO 👍

src/cli.rs
```diff
@@ -26,6 +28,10 @@ pub struct Args {
     #[clap(long, action)]
     /// Print version
     pub version: bool,
+
+    /// Config file to use
+    #[clap(short, long)]
+    pub config: Option<PathBuf>
```
altsem, line 34: This is a separate feature. I'd pull it out as a separate commit if possible.
Just since the changelog is generated from commits: https://github.com/altsem/gitu/blob/master/docs/dev-tooling.md

altsem (changes requested):
Just thinking about the commit message (and generated changelog).
Perhaps it'd be nice to communicate a bit more:
- feat(config): Removed support of sequences of keys `abc`
- feat(config): Removed requirement of `<>` around special keys
- fix(config): Print error an error message when a bound key is invalid

or something like that. What do you think?
It may be a pain to split the PR into these commits, adding empty commits is fine too imo.
As long as you prefix it with `feat` or `fix` then there'll be a changelog entry.

In the meantime I'll ponder whether changelog generation is really worth it :clown_face:

altsem:
I'm just rolling with the latest version of clippy in CI, so it breaks sometimes whenever new linting rules are added. I pushed a fix for it in master, so if you rebase, it should fix it.

cjordan:
Thanks for the review! I'll try to do everything indicated soon.

Also I've just looked at ctrl+j again; when I hit that, `gitu` reports a `KeyEvent` of "Enter". I tried a different terminal and got the same thing. Maybe my shell is somehow mapping this? Will investigate more soon.

cjordan:
It appears that on linux terminals, `ctrl+j` is a line feed and `ctrl+m` is a carriage return. Will keep investigating.

src/cli.rs
cjordan: Looks like this has independently gone into `master`, so it disappeared from my branch on rebase.

altsem:
@cjordan No problem! I think the commit situation is fine, it's just to produce a changelog which I'd do either way.

I think some terminals treat ctrl+j differently, so it might've been a mistake binding it by default actually. Not sure how the landscape looks like tbh.

I'll have another look :)

alexrp:
I tried the obvious `"space"` but that didn't work FWIW.

alexrp:
#449

PR #420, opened 2025-09-19
feat: Show diff on the stash detail screen

jonathanj:
Currently, the stash detail screen shows "unmerged" for all the files in the stash, which also means there is also no diff being shown for the stash.

It seems like the reason for "unmerged" is that the current implementation does `git show <stash_commit>` which ends up with `diff --cc` output, which the diff parser interprets as conflicts.

What I've changed here is to have a separate `show_stash` screen and to store the stash ref instead of the commit on the `ItemData::Stash` struct. The new screen runs `git stash show -p <stash_ref>` instead, which produces a diff that the diff parser is happier about.

I thought it was better to add something explicitly for showing stashes, rather than to try have a way to conditionally modify the diff parser, since it separates the concerns better at the cost of some conciseness in the code. More than happy to take suggestions.

### DIFF
--- a/src/git/mod.rs
+++ b/src/git/mod.rs
@@ -177,6 +177,24 @@ pub(crate) fn show(repo: &Repository, reference: &str) -> Res<Diff> {
     })
 }
 
+pub(crate) fn stash_show(repo: &Repository, stash_ref: &str) -> Res<Diff> {
+    let text = String::from_utf8(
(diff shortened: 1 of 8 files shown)


jonathanj:
Thank you @altsem! Sorry about the linter error, how do I run the linter locally for future reference?

PR #423, opened 2025-09-22
feat: revert to using Crossterm as backend (fixes rendering/input bugs)

altsem:
Started a bit on going back to using crossterm. I reverted the commits, but might've missed a bunch of stuff. it's not compiling yet. All of this could be squashed to a single commit I think. It's in a messy state.


altsem:
@jonathanj I did some progress, thought it was working great.
- No longer need `.saturating_sub(1)` when dealing with mouse events, they're 0-indexed
- Mouse capture/reporting is enabled by default in Crossterm. So I had to switch that.
- I don't encounter the strange key inputs when mouse capture is on.
- It seems to fix the issue of Ctrl+j being interpreted as enter (making hunk lines selectable again)
- Rendering issues when background color touches end of screen is no longer present.
- Achieved world peace

The tests are broken though D:

jonathanj:
I see you already modified the `mouse_event` to accept the crossterm values, so all I did was adjust the Y-position inside `mouse_event`. I would advocate for this because of how much it helps the test+snapshot iteration loop, over the technical correctness of the underlying library.

```diff
diff --git a/src/tests/helpers/ui.rs b/src/tests/helpers/ui.rs
index 4a75008..76dc5cd 100644
--- a/src/tests/helpers/ui.rs
+++ b/src/tests/helpers/ui.rs
@@ -136,7 +136,7 @@ pub fn mouse_event(x: u16, y: u16, mouse_button: MouseButton) -> Event {
     Event::Mouse(crossterm::event::MouseEvent {
         kind: crossterm::event::MouseEventKind::Down(mouse_button),
         column: x,
-        row: y,
+        row: y.saturating_sub(1),
         modifiers: KeyModifiers::NONE,
     })
 }
```

PR #424, opened 2025-09-24
Support file line positions for the micro editor

Profpatsch:
`micro` is a popular modern successor to `nano`
and uses the same `+line` syntax as `nano` and `vim`.


altsem (approved):
Nice!

PR #434, opened 2025-10-06
Fix nix builds with newer rust and source filters

ipsavitsky:
Currently, the nix builds are broken for two reasons:

1. Old version of the rust toolchain in an old version of fenix
2. Source filtering filtered out `.pest` sources when building

This change fixes both of those issues so that gitu is buildable with nix again


altsem (approved):
Thanks! I noticed I had broken these. Not a Nix user myself.

PR #445, opened 2025-10-29
feat: Implement spinoff branch command

bezbac:
I experimented with gitu a few months ago and noticed it didn't implement the spinoff branch functionality from magit, which I use quite heavily. Since I had previously implemented spinoff branch in edamagit, a magit extension for VS Code (https://github.com/kahole/edamagit/pull/290, https://github.com/kahole/edamagit/pull/294), I thought I'd give it a shot here as well.

I tried to match the code style and ensure there are no clippy issues, but if there's anything I should change, just let me know! :)

### DIFF
--- a/src/ops/branch.rs
+++ b/src/ops/branch.rs
@@ -165,9 +165,7 @@ impl OpTrait for Spinoff {
             }
 
             if does_branch_exist(&app.state.repo, &new_branch_name).unwrap_or(false) {
-                app.display_error(format!(
-                    "Cannot spin off {new_branch_name}. It already exists"
-                ));
+                return Err(Error::SpinoffBranchExists(new_branch_name.to_string()));
             }
 
             let Ok(current_branch) = get_current_branch(&app.state.repo) else {
--- a/src/error.rs
+++ b/src/error.rs
@@ -53,6 +53,7 @@ pub enum Error {
     PromptAborted,
     NoMoreEvents,
     CannotSpinoffCurrentBranch,
+    SpinoffBranchExists(String),
     DoesBranchExist(git2::Error),
     GetBranchName(git2::Error),
 }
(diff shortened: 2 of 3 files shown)


altsem:
I like! I'll try find some time to review this. But looks pretty good at a glance.

src/ops/branch.rs
```diff
@@ -133,3 +136,113 @@ pub fn delete(app: &mut App, term: &mut Term, branch_name: &str) -> Res<()> {
...
+            )?;
+
+            if new_branch_name.is_empty() {
+                return Err(Error::BranchNameRequired);
+            }
+
+            if does_branch_exist(&app.state.repo, &new_branch_name).unwrap_or(false) {
+                app.display_error(format!(
+                    "Cannot spin off {new_branch_name}. It already exists"
+                ));
```
altsem: app.display_error happens if you return an Err too. Which is what I've gravitated towards elsewhere.

I've put all these in src/error.rs. To pass data in data, you could do like:
```rust
return Err(Error::SpinoffBranchExists(new_branch_name))
```
and
```rust
// In error.rs
    SpinoffBranchExists(String),
```

src/ops/branch.rs
```diff
@@ -133,3 +136,113 @@ pub fn delete(app: &mut App, term: &mut Term, branch_name: &str) -> Res<()> {
...
+
+            let Some(upstream_branch_commit_oid) = upstream_branch_commit_oid else {
+                app.display_info(format!("Branch {current_branch_name} not changed"));
+                return Ok(());
+            };
+
+            if base_commit_oid == upstream_branch_commit_oid {
+                app.display_info(format!("Branch {current_branch_name} not changed"));
+                return Ok(());
+            }
```
altsem: Is it redundant to print that nothing was done? 🤔

bezbac: I think there's value in keeping this feedback. Since spinoff might change the branch it was started from (depending on whether the commits were local or already pushed to the remote), users might not always have the current branch state or commit history top of mind before running the command. This additional feedback helps clarify what actually happened.

The info message is the the same in [edamagit](https://github.com/kahole/edamagit/blob/61b9afeeacf4cb41491d9646d187f7702a7e2fb9/src/commands/branchingCommands.ts#L260-L263), but apparently not in the [original magit](https://github.com/magit/magit/blob/3d1a008f1894dd087aee099629e3726877956033/lisp/magit-branch.el#L491-L530).

bezbac:
Thanks for the feedback! I've adjusted the code accordingly. Please let me know if there's anything else that would improve the PR.

PR #467, opened 2025-12-07
fix: config syntax_highlight.enabled not working

gotschmarcel:
The config flag was simply not used.


gotschmarcel:
Didn't expect a fix so quick. You beat me to it :)

PR #468, opened 2025-12-10
feat: add interactive picker

sei40kr:
## Summary

This PR adds an interactive fuzzy finder picker component for branch and commit selection:

- **Fuzzy matching** with real-time filtering
- **Keyboard navigation** (↓/↑, Ctrl-n/p)
- **Select/Cancel** with Enter/Esc/Ctrl-c
- **Visual highlighting** of matched characters

## Implementation Details

- New `picker` module with fuzzy matching logic
- `ui/picker` for rendering the picker interface
- Configurable keybinds and styling via config.toml
- Test coverage for core picker functionality

## Test Plan

- [x] Unit tests (only for state)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,1 @@
+/nix/store/5jr06kaid73ilblgxp0njd5k35nxynnx-pre-commit-config.json
--- a/src/tests/snapshots/gitu__tests__non_utf8_diff.snap
+++ b/src/tests/snapshots/gitu__tests__non_utf8_diff.snap
@@ -1,0 +1,25 @@
+---
+source: src/tests/mod.rs
+expression: ctx.redact_buffer()
+---
+▌On branch main                                                                 |
+▌Your branch is ahead of 'origin/main' by 1 commit(s).                          |
+                                                                                |
+ Unstaged changes (1)                                                           |
+ modified   non_utf8.txt                                                        |
+ @@ -1 +1 @@                                                                    |
+ -File with valid UTF-8                                                         |
+ \ No newline at end of file                                                    |
+ +FileFile with invalid UTF-8: ��                                               |
+                                                                                |
+ Recent commits                                                                 |
+ 7c3d61a main add non_utf8.txt                                                  |
+ b66a0bf origin/main add initial-file                                           |
+                                                                                |
+                                                                                |
+                                                                                |
+                                                                                |
+                                                                                |
+                                                                                |
+                                                                                |
+styles_hash: 232c1529ecd86f60


src/ui/picker.rs
```diff
@@ -0,0 +1,142 @@
...
+    layout.horizontal(None, OPTS, |layout| {
+        let status_text = if state.loading {
+            " Loading...   ".into()
+        } else {
+            let current = if state.filtered_count() == 0 {
+                0
+            } else {
+                state.cursor() + 1
+            };
+            format!(" {}/{}   ", current, state.filtered_count()).into()
```
sei40kr:
```rust
            format!(" {}/{}   ", state.filtered_count(), state.total_items()).into()
```

I thought I had done it like this...

src/ui/picker.rs
```diff
@@ -0,0 +1,142 @@
...
+        layout_span(layout, (state.input_state.value().into(), Style::new()));
+        layout_span(layout, (CARET.into(), Style::new()));
+    });
+
+    // Calculate visible items range (scroll window)
+    let cursor = state.cursor();
+    let total = state.filtered_count();
+    let visible_count = MAX_ITEMS_DISPLAY.min(total);
+
+    let (start, _end) = if total <= MAX_ITEMS_DISPLAY {
```
sei40kr: Since `end` isn't used, we don't need to calculate it here.

src/ui/picker.rs
```diff
@@ -0,0 +1,142 @@
...
+            };
+            format!(" {}/{}   ", current, state.filtered_count()).into()
+        };
+        layout_span(layout, (status_text, info_style));
+
+        // Prompt with separator (like regular prompt)
+        layout_span(layout, (state.prompt_text.as_ref().into(), prompt_style));
+        layout_span(layout, (" › ".into(), Style::new().cyan().dim()));
+        layout_span(layout, (state.input_state.value().into(), Style::new()));
+        layout_span(layout, (CARET.into(), Style::new()));
```
sei40kr: I want to change the caret shape to indicate that it is in input mode.

src/ui/picker.rs
```diff
@@ -0,0 +1,142 @@
...
+        // Center the cursor in the visible window
+        let half = MAX_ITEMS_DISPLAY / 2;
+        if cursor < half {
+            (0, MAX_ITEMS_DISPLAY)
+        } else if cursor >= total - half {
+            (total - MAX_ITEMS_DISPLAY, total)
+        } else {
+            (cursor - half, cursor + half + (MAX_ITEMS_DISPLAY % 2))
+        }
+    };
```
sei40kr: It might be better to extract this into a separate function.

altsem:
Do you think a tab/shift-tab keybinding would be a nice default in addition to ctrl-n. ctrl-p?

src/ui/picker.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::picker::PickerState;
+use crate::ui::layout::OPTS;
+use crate::ui::{layout_span, repeat_chars, UiTree, CARET, STYLE};
+
+const MAX_ITEMS_DISPLAY: usize = 10;
+
+/// Layout the picker UI
+pub(crate) fn layout_picker<'a>(
+    layout: &mut UiTree<'a>,
+    state: &'a PickerState,
```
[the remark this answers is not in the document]
altsem: Could be worth testing the picker in a large repository where you'd get a lot of matches. To make sure the UI doesn't freeze.

sei40kr:
@altsem 
> Do you think a tab/shift-tab keybinding would be a nice default in addition to ctrl-n. ctrl-p?

I think that's a good idea. Especially for users familiar with Magit, many are likely accustomed to using Tab/Shift-Tab for navigation.

.pre-commit-config.yaml
altsem: Leftover?

src/default_config.toml
```diff
@@ -227,3 +232,8 @@ stash_menu.stash_pop = ["p"]
 stash_menu.stash_apply = ["a"]
 stash_menu.stash_drop = ["k"]
 stash_menu.quit = ["q", "esc"]
+
+picker.next = ["down", "ctrl+n"]
+picker.previous = ["up", "ctrl+p"]
```
altsem: I believe the tab bindings would be like this.
```suggestion
picker.next = ["down", "ctrl+n", "tab"]
picker.previous = ["up", "ctrl+p", "shift+tab"]
```

altsem (approved):
Actually, I'll just merge this to master. And make the small adjustments i commented on. I think it looks pretty good!
I suppose in the other PR, things will get more interesting when it is actually put to use!

PR #476, opened 2025-12-13
feat: align menu key and label spacing

sei40kr:
Improve menu display by aligning keys and labels in columns. Keys are right-padded to match the longest key in each column.

When multiple key bindings are grouped for the same operation (e.g., "k/up" or "ctrl+k/ctrl+up"), the maximum width is calculated based on the combined string length.

Before:
```
 Y Show Refs              b Branch
 k/up Up                  c Commit
 ctrl+k/ctrl+up Up line   h/? Help
```

After:
```
 Y              Show Refs   b   Branch
 k/up           Up          c   Commit
 ctrl+k/ctrl+up Up line     h/? Help
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


altsem (approved):
"fix: update Line iteration to use IntoIterator" would be in the changelog due to the "fix: " prefix, which it likely should not. Looks good otherwise!

PR #491, opened 2026-01-02
feat: add scroll view up/down, without moving selection

sbillig:
This separates scrolling from item selection, so the view can be scrolled down without moving the selection down. This is useful when viewing large change hunks that don't fit within the terminal viewport.

The default keybindings are ctrl+y to scroll up, ctrl+e to scroll down. Mouse/trackpad scrolling works well with the config setting `mouse_support = true`.

## Checklist


altsem:
Been holding off on this for too long and thinking what the default should be. But the pain of the cursor moving maybe outweighs the pain of moving the cursor 500 lines down after having scrolled so far down (I can't recall a single time I even did this anyway). In addition, anyone could just click their mouse on the line they want to jump to at that point :clown_face:.

I'll just fix these conflicts and get it merged!

PR #523, opened 2026-04-04
feat: log: search changes

PRESFIL:
## Checklist

### DIFF
--- a/src/items.rs
+++ b/src/items.rs
@@ -7,6 +7,7 @@ use crate::highlight;
 use crate::item_data::ItemData;
 use crate::item_data::Ref;
 use crate::item_data::SectionHeader;
+use git2::DiffOptions;
 use git2::Oid;
 use git2::Repository;
 use ratatui::text::Line;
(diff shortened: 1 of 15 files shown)


src/items.rs
```diff
@@ -350,6 +352,43 @@ pub(crate) fn log(
...
+                let parent_tree = parent
+                    .as_ref()
+                    .map(|p| p.tree())
+                    .transpose()
+                    .map_err(Error::ReadLog)?;
+                let current_tree = commit.tree().map_err(Error::ReadLog)?;
+
+                let mut opts = DiffOptions::new();
+                let diff = repo
+                    .diff_tree_to_tree(parent_tree.as_ref(), Some(&current_tree), Some(&mut opts))
```
altsem, line 366: I reckon this becomes quite expensive, diffing each commit to the previous. It does take a few seconds to complete a search for me in the Gitu repo with a release-build.

I've been replacing other parts of (lib)git2 uses with simply command forks. `git log --oneline --grep 'test'` might do the job. Perhaps there's a lower-level, more stable command than `git log --grep` that is more suitable for scripting (aka plumbing-command). Would require some rework.

PR #539, opened 2026-06-09
fix: calculate display width with unicode-width to handle emoji

cedricpinson:
## Problem

String display width was computed with grapheme cluster count (`graphemes(true).count()`). A single emoji such as 🚀 is one grapheme cluster but occupies **two** terminal columns. As a result the layout underestimated the width by one column per emoji, shifting and overlapping the spans that follow on the same line.

## Fix

Use the actual terminal display width from the `unicode-width` crate wherever a string's rendered width is needed:

- `layout_span` in `src/ui.rs` (the central function used for rendering)
- the `text()` helper in `src/ui/layout/mod.rs`
- `span_width` (overflow detection) in `src/screen/mod.rs`, and the ellipsis truncation, which now accumulates **display width** instead of grapheme count so a wide character is never split and the `…` column is correctly reserved

`unicode-width` was already a dev-dependency (used in test helpers); it is moved to `[dependencies]` since it is now needed at runtime.

## Tests

- New `emoji_display_width` test: a single emoji measures 2 columns and the following span starts at column 2.
- Full suite passes (395 tests), clippy clean.


cedricpinson:
Hi,
I was using the tool but always had some lines cut because of how the length of lines was computed. This PR is done by me, but the code comes from Claude. It's not a lot of code, but it’s worth mentioning the source of the fix.

Thanks

altsem:
Thanks for the heads up @cedricpinson I have been working on a lot of changes tangent to the diff of this one, so i incorporated the changes on the branch i have.

